PR #1, opened 2021-04-16
Fixed parse_mode being wrongly placed into the reply_markup object

MegaITA:
I found out that in the docs you have put `parse_mode` inside the `reply_markup` object just like this:

```
await ctx.reply(textWithHtml, {
  parse_mode: "HTML", // in order to pass value for `parse_mode`
  reply_markup: {
    // have to create custom `reply_markup` object
    keyboard: keyboard.build(), // note the `build` call
  },
});

```

I fixed it by moving it into the `options` object, where it should be.


KnorpelSenf:
Good catch, LGTM. Thanks!

PR #3, opened 2021-05-13
Fixed Router Example Documentation for TS

Tecardo1:
I adapted the example to the new session style. I did this only for TS, as it's untested.


Tecardo1:
sorry guys... I wrote export type MyContext.

KnorpelSenf (approved):
Good catch to update this! I fixed some things so it executes. Thanks!

PR #21, opened 2021-09-28
Fixed wrong ctx.getFile() example

dzek69:
Fixed wrong `ctx.getFile()` example, see: https://t.me/grammyjs/17075


KnorpelSenf (changes requested):
Thanks!

I think the link to the guide is missing a leading `/`

dzek69:
You are right, fixed

KnorpelSenf (approved):
Awesome, thanks for fixing this

PR #48, opened 2021-11-02
fix: wrong resource links

WingLim:
Some links in /zh is wrong, fix it.


linbuxiao (approved):
Great! This looks like the aftermath of a global replacement that never caught up with us.

KnorpelSenf:
Sorry, that was my bad. Thanks for fixing it!

PR #50, opened 2021-11-02
feat: migrate to vuepress-vite

### LINKED ISSUE #49: feat: migrate vuepress to vuepress-vite or vitepress
`vuepress` is using webpack which is much slower than `vuepress-vite` or `vitepress`(they are using vite and esbuild).

I have tested the build time between `vuepress` and `vuepress-vite`, here is the result:

```
+---------+----------+---------------+
|         | vuerepss | vuepress-vite |
+---------+----------+---------------+
| Time1   | 30.29s   | 17.77s        |
| Time2   | 29.67s   | 16.87s        |
| Average | 29.98s   | 17.32s        |
+---------+----------+---------------+
```

`vuepress-vite` is **40%** faster than `vuepress` in my `i5-9500` computer.

Migrate to `vuepress-vite` is very simple and easy, or we can migrate to `vitepress`(to vitepress is a bit complicated, but I will do the job).

There will be a PR migrate to `vuepress-vite`, you can take a look about it.

WingLim:
Closes #49.

### DIFF
--- a/site/docs/.vuepress/config.ts
+++ b/site/docs/.vuepress/config.ts
@@ -1,5 +1,5 @@
-import { defineUserConfig } from "vuepress";
-import type { DefaultThemeOptions } from "vuepress";
+import { defineUserConfig } from "vuepress-vite";
+import type { DefaultThemeOptions } from "vuepress-vite";
 
 export default defineUserConfig<DefaultThemeOptions>({
   title: "grammY",
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -1,6106 +1,8 @@
 {
   "name": "grammy-website",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "grammy-website",
-      "version": "1.0.0",
-      "license": "MIT",
-      "devDependencies": {
-        "@vuepress/plugin-docsearch": "^2.0.0-beta.26",
-        "typescript": "^4.4.4",
-        "vuepress": "^2.0.0-beta.26"
-      }
-    },
-    "node_modules/@algolia/autocomplete-core": {
-      "version": "1.0.0-alpha.44",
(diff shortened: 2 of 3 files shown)


KnorpelSenf (changes requested):
LGTM!

Please update the docsearch dependency to beta.27 so that it's in sync with `vuepress-vite`.

WingLim:
@KnorpelSenf Already update docsearch to newest.

KnorpelSenf (approved):

KnorpelSenf:
@WingLim Why did this PR downgrade the lockfile version to `1`?

WingLim:
> @WingLim Why did this PR downgrade the lockfile version to `1`?

Oops, I check my mac using `pnpm` to manage node version.
`node` and `npm` switch back to old version, and I didn't notice it.
Will upgrade it back...

KnorpelSenf:
Yet another alternative would be to hard-code an exception for the respective paragraph in the `<wbr>` tag injection plugin. I am not too convinced that this is the right thing to do, but it would certainly work.

We could revert this change once https://github.com/vuepress/vuepress-next/issues/602 is fixed.

KnorpelSenf:
I believe it's only the single paragraph that I linked. I read the complete docs.

WingLim:
> I believe it's only the single paragraph that I linked. I read the complete docs.

Let's see what's @roj1512 option, I prefer the second, `vite` is faster for dev in preview.

roj1512:
@KnorpelSenf I am not seeing any duplicate text on the guide/basics page when I run locally.

KnorpelSenf:
That's part of the bug. This behaviour is described in the aforementioned issue.

roj1512:
How should it get rephrased btw? Get the text out of quotes?

KnorpelSenf:
I don't think it's happening only in the quotes, does it? We'll just have to play a bit around with it until it suddenly doesn't appear anymore. That is, unless you want to understand the bug, but at that point you might as well fix vuepress.

PR #52, opened 2021-11-02
fix: link path and menu translation

WingLim:
Still some link is wrong in nav bar, and `Telegram` no need to be translated.


linbuxiao (approved):
nice！

PR #64, opened 2021-11-03
chore: upgrade package-lock.json to version 2

WingLim:
This is because an older version of npm was used to generate this file.
Fix it by using npm `v8.1.0`.
Sorry for the distress.


KnorpelSenf (approved):
No worries, happens to the best of us. Thanks for the prompt fix!

PR #79, opened 2021-11-07
Add semicolon to pass CI

WingLim:
Fail in CI with missing semicolon.


KnorpelSenf (approved):
I think this is already contained somewhere, but I lost track. Won't hurt to merge. Thanks!

PR #81, opened 2021-11-08
Add menu plugin docs

KnorpelSenf:
Title says it all.

@CalsiBotDev would you mind having a look?

@EdJoPaTo @KnightNiwrem @backmeupplz you may be interested in how the plugin actually works. This is the first proper explanation of what is going on. [Check the Netlify deploy preview](https://6188a55adb065100089fdeef--grammy.netlify.app/plugins/menu) from the CI below. Feedback is welcome!

@WingLim this is a lot to translate. Leave a comment if we should wait for you with merging this. Otherwise I will assume that we just merge English first, and receive a Chinese version eventually in a different PR.


WingLim:
This may take some time, I think you better merge it first, and I am staring now.

site/docs/plugins/keyboard.md
```diff
@@ -87,6 +87,13 @@ Specify an empty inline keyboard to remove all buttons underneath a message.
 
 ### Responding to Clicks
 
+::: tip Menu Plugin
+The keyboard plugin gives you raw access to the update objects that Telegram is sending.
+However, responding to clicks that way can be tedious.
+If you are looking for a more high-level implementation of inline keybaords, check out [the menu plugin](./menu.md).
```
KnightNiwrem, line 93: Spelling

site/docs/plugins/keyboard.md
```diff
@@ -87,6 +87,13 @@ Specify an empty inline keyboard to remove all buttons underneath a message.
...
+It makes it simply to create interactive menus.
```
KnightNiwrem, line 94: simple*

site/docs/plugins/keyboard.md
KnorpelSenf: What? I don't see it

KnightNiwrem: Ah.. search for `keybaords`.

WingLim:
@KnorpelSenf I have double check the translation, it should be well translated now.

KnorpelSenf:
Impressive, that was fast!

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+const menu = new Menu("id", { autoAnswer: false });
+```
+
+You will now have to call `answerCallbackQuery` yourself.
+This allows you to pass custom messages that are displayed to the user.
 
 ## Outdated Menus and Fingerprints
 
+Let's say you have a menu where a user can toggle notifications on and off, such as in the example [up here](#dynamic-labels).
+Now, if a user sends e.g. `/settings` twice, they will get the same menu twice.
```
CalsiBotDev, line 380:
```suggestion
Now, if a user sends `/settings` twice, they will get the same menu twice.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+But changing the notification setting on one of the two messages will not update the other!
```
CalsiBotDev, line 381:
```suggestion
But, changing the notification setting on one of the two messages will not update the other!
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+
+It is clear that we cannot keep track of all settings messages in a chat, and update all old menus in the entire chat history.
```
CalsiBotDev, line 383:
```suggestion
It is clear that we cannot keep track of all settings messages in a chat, and update all old menus across the entire chat history.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+You would have to use so many API calls for this that Telegram will restrict your bot.
```
CalsiBotDev, line 384:
```suggestion
You would have to use so many API calls for this that Telegram would rate-limit your bot.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+You also need to buy a lot of storage to remember all the message identifiers of all menus in all chats.
```
CalsiBotDev, line 385:
```suggestion
You would also require a lot of storage to remember all of the message identifiers of every menu, across all chats.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+That is not practical.
```
CalsiBotDev, line 386:
```suggestion
This is not practical.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+
+The solution to this is to check if a menu is outdated before performing any actions.
```
CalsiBotDev, line 388:
```suggestion
The solution, is to check if a menu is outdated _before_ performing any actions.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+That means that we will only update old menus if the user actually starts clicking any buttons on them.
+The menu plugin does this automatically for you, so you don't need to worry about it.
```
CalsiBotDev, line 390:
```suggestion
The menu plugin handles this automatically for you, so you don't need to worry about it.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+Now, if a user sends e.g. `/settings` twice, they will get the same menu twice.
+But changing the notification setting on one of the two messages will not update the other!
+
+It is clear that we cannot keep track of all settings messages in a chat, and update all old menus in the entire chat history.
+You would have to use so many API calls for this that Telegram will restrict your bot.
+You also need to buy a lot of storage to remember all the message identifiers of all menus in all chats.
+That is not practical.
+
+The solution to this is to check if a menu is outdated before performing any actions.
+That means that we will only update old menus if the user actually starts clicking any buttons on them.
```
CalsiBotDev, line 389:
```suggestion
This way, we will only update old menus if a user actually starts clicking the buttons on them.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+The menu plugin does this automatically for you, so you don't need to worry about it.
+
+You can configure what happens when an outdated menu is detected.
```
CalsiBotDev, line 392: Since there is full control, and not a choice between presets.
```suggestion
You can configure exactly what happens when an outdated menu is detected.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+    await ctx.reply("Here is a fresh menu", { reply_markup: menu1 });
+  },
+});
+// Completely disable outdated check (may run wrong button handlers)
+const menu2 = new Menu("id", { onMenuOutdated: false });
+```
+
 We have a heuristic to check if the menu is outdated.
-Read the API reference to see config.
+It looks at
```
CalsiBotDev, line 411:
```suggestion
It looks at:
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+- the identifier of the menu,
+- the shape of the menu,
+- the position of the pressed button,
+- the payload, if specified, and
+- the text of the pressed button.
+
+This data is effeciently compressed into a 4-byte hash that is stored in every button.
+It will then be compared to see if the menu is outdated before any handler is run.
+
+It is thinkable that your menu changes, but all of the above things are staying the same.
```
CalsiBotDev, line 422:
```suggestion
It is possible that your menu changes, while all of the above things stay the same.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+This is usually not the case, but if you are creating a menu that has this property, you should use a fingerprint function.
```
CalsiBotDev, line 423:
```suggestion
This is usually not the case, but if you are creating a menu where this is likely, you should use a fingerprint function.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+
+```ts
+function ident(ctx: Context): string {
+  // return string that would change when your menu changes
+}
+const menu = new Menu("id", { fingerprint: (ctx) => ident(ctx) });
+```
+
+The fingerprint string will be added to the above list of things that impact the hash generation.
+That way, you can make sure that outdated menus are always detected.
```
CalsiBotDev, line 433:
```suggestion
This way, you can be sure that outdated menus are always detected.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+
+## How Does It Work
+
+The menu plugin works completely without storing any data.
+This is important for large bots with millions of users.
+Saving the states of all menus would consume too much memory.
```
CalsiBotDev, line 439:
```suggestion
Saving the state of all menus would consume too much memory.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+
+When you create your menu objects and link them together via `register` calls, no menus are actually built.
+Instead, the menu plugin will remember how to assemble new menus based on your operations.
+Whenever a menus is sent, it will replay these operation to render your menu.
```
CalsiBotDev, line 443:
```suggestion
Whenever a menu is sent, it will replay these operations to render your menu.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+This includes layouting all dynamic ranges and generating all dynamic labels.
```
CalsiBotDev, line 444:
```suggestion
This includes laying out all dynamic ranges and generating all dynamic labels.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+Once the menu is sent, the rendered button array will be forgotten again.
+
+When a user presses a button of the menu, we need to find the handler that was added to that button at the time the menu was rendered.
```
CalsiBotDev, line 447:
```suggestion
When a user presses a menu's button, we need to find the handler that was added to that button at the time the menu was rendered.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+Hence, we simply render the old menu again.
+However, this time, we don't actually need the full layout—all we need is the overall structure, and that one specific button.
+Consequently, the menu plugin will perform a shallow rendering in order to be more efficient.
+In other words, the menu will only be rendered partially.
+
+Once the pressed button is known again (and we have checked that the menu is not [outdated](#outdated-menus-and-fingerprints)), we invoke the handler.
+
+Internally, the menu plugin makes heavy use of [API Transformer Functions](/advanced/transformers.md), e.g. to quickly render outgoing menus on the fly.
```
CalsiBotDev, line 455:
```suggestion
Internally, the menu plugin makes heavy use of [API Transformer Functions](/advanced/transformers.md), for example, to quickly render outgoing menus on the fly.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
 
-TODO: explain the problem and the solution, mention shortcomings, explain config
+When you register the menus in a large hierachy for navigating, they will in fact not store these references explicitly.
```
CalsiBotDev, line 457:
```suggestion
When you register the menus in a large navigation hierarchy, they will in fact not store these references explicitly.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+Under the hood, all menus of that one structure are added to the same large pool, and that pool is shared across all contained instances.
+Every menu is responsible for every other one in the index, and they can handle and render each other.
+(Most often, it is only the root menu that is actually passed to `bot.use` and receiving any updates.
```
CalsiBotDev, line 460:
```suggestion
(Most often, it is only the root menu that is actually passed to `bot.use` and that receives any updates.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+In that case, this one instance will handle the complete pool.)
```
CalsiBotDev, line 461:
```suggestion
In such cases, this one instance will handle the complete pool.)
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+As a result, you are able to navigate between arbitrary menus without limits, and still the update handling can happen in `O(1)` because there is no need to search through hierarchies after the right menu to handle a button click.
```
CalsiBotDev, line 462: Not sure how ` are handled in hyperlinks, will this work?
```suggestion
As a result, you are able to navigate between arbitrary menus without limit, all while the update handling can happen in [`O(1)` time complexity](https://en.wikipedia.org/wiki/Time_complexity#Constant_time) because there is no need to search through entire hierarchies to find the right menu to handle any given button click.
```

site/docs/plugins/keyboard.md
```diff
@@ -87,6 +87,13 @@ Specify an empty inline keyboard to remove all buttons underneath a message.
 
 ### Responding to Clicks
 
+::: tip Menu Plugin
+The keyboard plugin gives you raw access to the update objects that Telegram is sending.
```
CalsiBotDev, line 91:
```suggestion
The keyboard plugin gives you raw access to the update objects that Telegram sends.
```

site/docs/plugins/keyboard.md
```diff
@@ -87,6 +87,13 @@ Specify an empty inline keyboard to remove all buttons underneath a message.
...
+However, responding to clicks that way can be tedious.
```
CalsiBotDev, line 92:
```suggestion
However, responding to clicks this way can be tedious.
```

CalsiBotDev (changes requested):

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+Naturally, if you are using a custom context type, you can pass it to `Menu`, too.
+
+```ts
+const menu = new Menu<MyContext>("id");
+```
 
 ## Adding Buttons
 
-Works like the inline keybaord plugin.
+The menu plugin layouts your keyboards exactly like [the plugin for inline keyboards](./keyboard.md#building-an-inline-keyboard) does.
```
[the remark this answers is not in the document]
KnorpelSenf: Oops

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+      .text(i.toString(), (ctx) => ctx.reply(`${i} selected`))
+      .row();
+  }
+  return range;
+});
+
+menu.text("Generate new", (ctx) => ctx.menu.update());
+```
+
+The range builder function that you pass to `dynamic` may be `async`, so you can even perform API calls or do database communication.
```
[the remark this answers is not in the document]
KnorpelSenf: This is not a menu! It's a menu range. I'm trying to make it very clear what the difference is. We could say “before returning the generated menu range”?

site/docs/plugins/menu.md
KnorpelSenf: Are you sure about the comma? Otherwise I agree

site/docs/plugins/menu.md
KnorpelSenf: Yes.

KnorpelSenf (commented):
Thank you for the thorough review! I applied most suggestions immediately. I also have a few comments.

KnorpelSenf:
Not sure why GitHub displays my comments twice. Have a look at your annotations, I replied that. You can ignore the comments displayed underneath my review.

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+The menu will detect that you edit the text of the message, and use the opportunity to update the buttons uderneath, too.
+As a result, you can often skip the `ctx.menu.update()` calls.
+
+Calling `ctx.menu.update()` does not update the menu immediately.
+Instead, it only sets a flag and remembers to update it at some point during your middleware execution.
+This is called _lazy updating_.
+If you edit the message itself later on, the plugin can simply use the same API call to also update the buttons.
+This is very efficient, and it makes sure that both the message and the keyboard are updated at the same time.
+
+Naturally, if you called `ctx.menu.update()` but you never perform any edits to the message, the menu plugin makes sure to update your buttons before you middleware completes.
```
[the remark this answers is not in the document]
CalsiBotDev: It refers to "the buttons" in this case, so "by themselves" is correct.

Maybe this is better though, and lets us use itself:
```suggestion
Naturally, if you call `ctx.menu.update()` but you never request any edits to the message, the menu plugin will update the keyboard by itself, before your middleware completes.
```

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
 ## Dynamic Ranges
 
-Use `menuA.dynamic` to add a dynamic number of buttons.
-Use a `MenuRange` for this.
+So far, we've only seen how to change the text on a button dynamically.
+You can also dynamically adjust the structure of a menu in order to add and remove buttons on the fly.
+
+::: danger Never Change a Menu During Message Handling
+You cannot create or change menus your during message handling.
+All menus must be fully created and registered before your bot starts.
```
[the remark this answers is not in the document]
CalsiBotDev: In that case, let's avoid the "should" and treat it as if it is not possible at all.

So, the first block of suggested changes.

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+:::
+
+You can let a part of a menu's buttons be generated on the fly (or all of them if you want).
+We call this part of the menu a _dynamic range_.
+The easiest way to create a dynamic range is by using the `MenuRange` class that this plugin provides.
+A `MenuRange` provides you with exactly the same functions as a menu, but it does not have an identifier, and it cannot be registered.
+
+```ts
+function getRandomInt(min: number, max: number) {
+  return min + Math.floor(Math.random() * (max - min));
```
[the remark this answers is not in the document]
CalsiBotDev: Oh, okay... I guess this is considered as "normal" so will be assumed to work that way?

To me it seems like really unexpected behavior. Maybe we can add a comment to avoid this issue:
```ts
// Returns a random number between and min [inclusive] and max [exclusive] 
```

site/docs/plugins/menu.md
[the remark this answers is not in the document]
CalsiBotDev: Yes, "menu range" works just as well here.

site/docs/plugins/menu.md
CalsiBotDev: The text is better with it, but will not be technically incorrect without it.

site/docs/plugins/menu.md
```diff
@@ -93,57 +93,373 @@ bot.start();
...
+```ts
+const menu = new Menu("pun-intended")
+  .text(
+    { text: "I know my creator", payload: (ctx) => ctx.from.first_name },
+    (ctx) => ctx.reply(`I was once created by ${ctx.match}!`),
+  );
+
+bot.use(menu);
+bot.command("menu", async (ctx) => {
+  await ctx.reply("You summoned a demenu!", { reply_markup: menu });
```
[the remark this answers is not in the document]
CalsiBotDev: Nope, I can't. 😔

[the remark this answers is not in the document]
KnorpelSenf: That's fine, I'll just remove it.

site/docs/plugins/menu.md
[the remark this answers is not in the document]
KnorpelSenf: Yes, that's default behaviour. Check python's `range`, and `String.prototype.substring` in JS, and all other things in all languages.
Note also that `Math.random()` itself returns values from `[0, 1)`. We just reflect that. In other words, if the 1 would be inclusive, `max` would automatically be inclusive, too.

I'm okay with adding a comment if that's helpful. I'm only hesitant because random numbers are not the point of the example.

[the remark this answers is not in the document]
KnorpelSenf: How about calling the arguments `minInclusive` and `maxExclusive`?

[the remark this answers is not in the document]
KnorpelSenf: Please have a look at 40a13a6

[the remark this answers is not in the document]
CalsiBotDev: That's perfect and much better than a comment.

I was originally going to suggest a change like this, but thought it would be less likely to be accepted than just adding a comment.

KnorpelSenf:
@WingLim English is done now. Sorry for all the changes. You might not need to adjust all translations, but that's rather to decide on a case-by-case basis.

Next time I'll only ping you after all changes so that you don't have twice the work.

KnorpelSenf:
Thanks! Looks ready to merge. Speak up everyone if you have objections! I'll merge and and release 1.0 later today.

WingLim:
@linbuxiao Maybe you want to take a look the new Chinese menu documentation here too.
Check if there are any mistakes.

PR #83, opened 2021-11-09
docs: translate zh/plugins/hydrate

WingLim:
- [x] hydrate.md

### DIFF
--- a/site/docs/plugins/hydrate.md
+++ b/site/docs/plugins/hydrate.md
@@ -1,4 +1,4 @@
-# Hydration Plugin for grammY
+# Hydration Plugin for grammY (`hydrate`)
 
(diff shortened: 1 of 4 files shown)


site/docs/plugins/hydrate.md
```diff
@@ -1,4 +1,4 @@
-# Hydration Plugin for grammY
+# Hydration Plugin for grammY (`hydrate`)
```
KnorpelSenf, line 1: Nice catch, thank you!

WingLim: Same to files plugin, I add it in #86 too.

linbuxiao (approved):
`hydrate`不翻译是正确的哈哈，nice！

PR #102, opened 2021-11-18
Fix typo in the holy name

rojserbest:


KnorpelSenf (approved):
Thanks!

rojserbest:
@KnorpelSenf, should we replace all `Node`s with `Node.js`?

KnorpelSenf:
I do not think this is necessary. Both forms are used a lot, even by the Node(.js) team themselves. I already put an explanation here: https://grammy.dev/guide/introduction.html#prerequisites-to-getting-started

> “If you see people call [Node.js] Node, they are just too lazy to type ‘.js’, but they mean the same thing.”

rojserbest:
I just thought it's uncool to have both variants on a single page.

KnorpelSenf:
I don't mind, but I also don't have strong opinions on the issue. You can change it to be consistent if you want :)

PR #110, opened 2021-11-28
Fix typo

deptyped:


KnorpelSenf (changes requested):
Have a look at https://github.com/grammyjs/website/pull/81#discussion_r745001667. I think we should rather rename the variables inside the function body.

KnorpelSenf:
Okay well it seems like GitHub doesn't allow me to link to closed conversations. The point is: someone was confused about the whole [inclusive, exclusive) convention, so we tried to name the variables explanatory.

KnorpelSenf (approved):
LGTM after the formatting script was run

KnorpelSenf:
Thanks!

PR #112, opened 2021-11-28
Rename code example tab names

KnorpelSenf:
Renames tab titles of code examples from TS to TypeScript. Does the same for JS and JavaScript.

This

- improves readability
- makes it easier to use the site on mobile (>15 % of page visits) because the buttons aren't so tiny
- makes it clearer to beginners who are not yet familiar with the abbreviations
- doesn't look as lazy

What do you think? Should we merge this?

### DIFF
--- a/site/docs/README.md
+++ b/site/docs/README.md
@@ -35,7 +35,7 @@ Bots are written in [TypeScript](https://www.typescriptlang.org) (or JavaScript)
 `npm install grammy` and paste the following code:
 
 <CodeGroup>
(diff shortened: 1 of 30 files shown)


KnorpelSenf:
You may want to have a look at the Deploy Preview: https://deploy-preview-112--grammy.netlify.app/

KnightNiwrem (approved):

WingLim (approved):
LGTM 🚀

PR #98, opened 2021-11-14
docs: Chinese documentation optimisation

WingLim:
Still, there are some format style issue, some words issue and more.
I am optimising the document to let Chinese reader have better experience.
This may take some time to do it. And need @linbuxiao your help here.

## Format
- [x] No space before or after the Chinese punctuation
- [x] Add space for link
- [x] Use Chinese brackets only for the content

## Translation
- [x] Better and easier translation to understand

PR #113, opened 2021-11-28
Add missing await

KnorpelSenf:
Title says it all


KnorpelSenf:
@WingLim please add this change to #98

KnorpelSenf:
Closed because based off of #112 for no reason.

PR #116, opened 2021-12-01
Rearrange examples

KnorpelSenf:
Currently, there are examples in three places.

1. Examples bots repo
2. Awesome grammY list
3. Community showlounge

The latter two are more or less identical. It does not make much sense and is mostly organised this way for historical reasons.

This PR suggests to get rid of the community showlounge, and to place the Awesome list more prominently in the Examples tab instead of in the resources. It will now be organised like this.

1. Awesome grammY list
2. Example bots repo

In the future, there are two more things that would be cool to have.

1. Example bots right inside the website. The ones in the repo are basically just a number of files that illustrate how to do certain operations, such as sending a message, or running a local bot API server in a Docker container. The examples on the website should rather be self-contained bots that are cool to play around with. They should feature
   - A Deno Deploy button to instantly deploy the code
   - Not much more than 50 lines of code
   - an illustration of some nifty grammY feature
   - a decent user experience when tested out in Telegram
   
   These bots must all have their own repo AFAIK. That means we should not have more than three of them to keep the maintenance low.
2. Live Browser Demo right inside the website. https://github.com/grammyjs/grammY/pull/111 made it trivial to turn the user's browser into a web server that runs a grammY bot. As a result, we can have a simple echo bot right there.

   It would also make a lot of sense to have a step-by-step wizard to set them up. Something like:
	1. Get bot token (information about how this will not be shared, a text field to paste it, and a small info tile that immediately displays the bot name once the token is inserted
	2. Pick a bot to run, just a list of bots, and a run button for each. Code should also be somewhere.
	3. TRY IT button that opens the bot's chat in Telegram (open a new tab with `t.me/bot_name`)

   That way, it should be super trivial to get your first bot up and running in your browser. Token can be stored in localstorage so that it is kept across navs and reloads.

   This could certainly use some community contributions from UI people :)


KnorpelSenf:
Now that I'm thinking about it, we may want to merge the two suggestions for the future. We could have three bots, and ech of them has a button to run it (will start the bot in the browser and immediately switch over to the TG client) and a second button right next to it to deploy it to Deno Deploy (akin to https://github.com/discordeno/serverless-deno-deploy-template#serverless-template)

rojserbest:
I really love these optimizations!

rojserbest (approved):

PR #122, opened 2021-12-03
Added Deta Base to the Adapter list

dcdunkan:
- Added [Deta.sh Base storage adapter](https://github.com/grammyjs/storage-deta) to the Official Adapter list
- Changed `then` to `than` --- seems like a typo.


KnorpelSenf (approved):
Thanks!

PR #130, opened 2021-12-06
Add section with common errors to FAQ

KnorpelSenf:
Requested by @darvesh in https://t.me/grammyjs/26591 and by @Tecardo1 in https://t.me/grammyjs/25116.

PR #132, opened 2021-12-06
Add entitiy parsing to FAQ

KnorpelSenf:
Continues #130.

PR #133, opened 2021-12-06
Sync changes to Chinese from #130

WingLim:
Releated #130.
Title says all.


KnorpelSenf:
Thanks! Have you seen #132? Do you want to integrate it here?

WingLim:
> Thanks! Have you seen #132? Do you want to integrate it here?

I think we could merge this first, and then I continue my work in #132.

KnorpelSenf (approved):
Sounds good!

PR #129, opened 2021-12-06
Add Deno Deploy guide

rojserbest:

PR #140, opened 2021-12-07
Translate deno-deploy.md to Chinese

WingLim:
- deno-deploy.md

Releated to #129

### DIFF
--- a/site/docs/zh/hosting/deno-deploy.md
+++ b/site/docs/zh/hosting/deno-deploy.md
@@ -1,4 +1,4 @@
-# 托管：Deno deploy
+# 托管：Deno Deploy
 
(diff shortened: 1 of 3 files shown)


site/docs/zh/hosting/deno-deploy.md
```diff
@@ -0,0 +1,59 @@
+# 托管：Deno deploy
```
WingLim, line 1: Maybe we should use `Deno Deploy` in the title?
@KnorpelSenf

KnorpelSenf: We definitely should, you're right!

KnorpelSenf (approved):
LGTM after the title update in both languages.

PR #142, opened 2021-12-08
Use GH Actions instead of Netlify to build the site

KnorpelSenf:
We only have a limited number of build minutes per month on Netlify. Too many updates to the site will exceed this limit.

With this PR, we can instead build the site on GH and push the assets to Netlify, which is free.


habemuscode (approved):
Good job!

KnorpelSenf:
I can confirm that works, I can see on Netlify that it has performed a production deploy for https://grammy.dev

KnorpelSenf:
And it also didn't consume any build minutes! 🥳

PR #147, opened 2021-12-11
Remove non-existing 'showlounge.md' next page

dcdunkan:
The file `docs/demo/showlounge.md` does not exists in both translations. And currently it looks like,
Screenshot_2021-12-11-20-51-46

### DIFF
--- a/site/docs/demo/examples.md
+++ b/site/docs/demo/examples.md
@@ -1,6 +1,5 @@
 ---
 prev: ./
-next: ./showlounge.md
 ---
 
 # Example Bots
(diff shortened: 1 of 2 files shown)


KnorpelSenf (approved):
Thanks! Missed the links when I removed the page in #116

KnorpelSenf:
@dcdunkan note that you are a member of the @grammyjs org since you started publishing an official storage adapter. Therefore, you have write permission to feature branches on all repositories. You don't need to have a fork of this repository—just branching and pushing directly to this repo works for you. That way, it reuses the Netlify secrets from the repo, and gives us automatic deployment previews. (Not saying we would've needed that here, but in the future that may happen.)

dcdunkan:
Oh, didn't know that :)
I'll do it in that way from now on.

PR #151, opened 2021-12-12
Improve code example of message sender type detection

KnorpelSenf:
Requested by @KnightNiwrem in https://t.me/grammyjs/29448


KnightNiwrem (approved):

KnorpelSenf:
I think that we do not have to wait for the Spanish translations yet. That would only make sense once Spanish is done and merged into `main`, as the target branch currently differs between the languages.

PR #148, opened 2021-12-12
[Fix] [Translations] [English]: Syntax fix and format tip section

habemuscode:

PR #161, opened 2021-12-13
fix: Do not use two English branches

KnorpelSenf:
I do not understand how this makes sense, so I suggest to drop having two parallel versions of English. Follow-up for https://github.com/grammyjs/website/pull/148#issuecomment-992276466


rojserbest (approved):

habemuscode (approved):

roj1512:
Thanks for merging, @habemuscode!

PR #162, opened 2021-12-13
feat: Keyboard to custom keyboard renaming and improvements

KnorpelSenf:
Suggested in https://t.me/grammyjs/29677

### DIFF
--- a/site/docs/plugins/keyboard.md
+++ b/site/docs/plugins/keyboard.md
@@ -16,7 +16,7 @@ Let us try to clear it up a bit:
 | **`Keyboard`**                           | class in grammY to create keyboards                                                                                               |
 | **`InlineKeyboard`**                     | class in grammY to create inline keyboards                                                                                        |
 
-> Note that both keyboard buttons and inline keyboard buttons can also have other functions, such as requesting the user location, opening a website, and so on.
+> Note that both keyboard buttons and inline keyboard buttons can also have other functions, such as requesting the user's location, opening a website, and so on.
 > This was omitted for brevity.
 
 ## Inline Keyboards
@@ -299,7 +299,7 @@ Moreover, buttons will always send exactly the message that's written on them.
 Telegram does not allow you to create buttons that display one text, but send another.
 If you need to do this, you should use an inline keyboard instead.
 
-In order to handle the click of a specific button, you can use `bot.hears` with the same text as you put on the button
+In order to handle the click of a specific button, you can use `bot.hears` with the same text as you put on the button.
 If you want to handle all button clicks at once, you use `bot.on('message:text')` and inspect `ctx.msg.text` to figure out which button was clicked, or if an ordinary text message was sent.
 
 ### Removing a Keyboard
--- a/site/docs/plugins/keyboard.md
+++ b/site/docs/plugins/keyboard.md
@@ -1,6 +1,23 @@
 # Keyboards and Inline Keyboards (built-in)
 
 Your bot may send a number of buttons, either to be [displayed underneath a message](#inline-keyboards), or to [replace the user's keyboard](#keyboards).
+They are called _inline keyboards_ and _keyboards_, respectively.
+If you think that this is confusing, then that's because it is.
+Thank you, Telegram, for this overlapping terminology.
+
+Let us try to clear it up a bit:
+
+| Term                                     | Definition                                                                                                                        |
+| ---------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| [**Keyboard**](#keyboards)               | a set of buttons that is displayed instead of the user's system keyboard                                                          |
+| [**Inline Keyboard**](#inline-keyboards) | a set of buttons that is displayed underneath a message inside the chat                                                           |
+| **Keyboard button**                      | a button in a keyboard, sends a text message with its label when pressed                                                          |
+| **Inline keyboard button**               | a button in an inline keyboard, sends an callback query not visible to the user when pressed, sometimes just called inline button |
+| **`Keyboard`**                           | class in grammY to create keyboards                                                                                               |
+| **`InlineKeyboard`**                     | class in grammY to create inline keyboards                                                                                        |
+
+> Note that both keyboard buttons and inline keyboard buttons can also have other functions, such as requesting the user location, opening a website, and so on.
+> This was omitted for brevity.
 
 ## Inline Keyboards
 
--- a/site/docs/plugins/keyboard.md
+++ b/site/docs/plugins/keyboard.md
@@ -10,7 +10,7 @@ Let us try to clear it up a bit:
 | Term                                     | Definition                                                                                                                          |
 | ---------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
 | [**Inline Keyboard**](#inline-keyboards) | a set of buttons that is displayed underneath a message inside the chat                                                             |
-| [**Custom Keyboard**](#custom-keyboards)        | a set of buttons that is displayed instead of the user's system keyboard                                                            |
+| [**Custom Keyboard**](#custom-keyboards) | a set of buttons that is displayed instead of the user's system keyboard                                                            |
 | **Inline keyboard button**               | a button in an inline keyboard, sends an callback query not visible to the user when pressed, sometimes just called _inline button_ |
 | **Custom Keyboard button**               | a button in a keyboard, sends a text message with its label when pressed, sometimes just called _keyboard button_                   |
 | **`InlineKeyboard`**                     | class in grammY to create inline keyboards                                                                                          |
@@ -287,9 +287,9 @@ You can specify the `selective` option if you want to show the custom keyboard o
 
 ```ts
 await ctx.reply(text, {
+  reply_to_message_id: ctx.msg.message_id,
   reply_markup: {
     selective: true,
-    reply_to_message_id: ctx.msg.message_id,
     keyboard: keyboard.build(),
   },
 });
--- a/site/docs/plugins/keyboard.md
+++ b/site/docs/plugins/keyboard.md
@@ -1,22 +1,22 @@
-# Keyboards and Inline Keyboards (built-in)
+# Inline and Custom Keyboards (built-in)
 
-Your bot may send a number of buttons, either to be [displayed underneath a message](#inline-keyboards), or to [replace the user's keyboard](#keyboards).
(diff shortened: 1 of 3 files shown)


site/docs/plugins/keyboard.md
```diff
@@ -273,6 +273,17 @@ await ctx.reply(text, {
 });
 ```
 
+### Responding to Clicks
+
+As mentioned earlier, all that keyboards do is sending regular text messages.
```
EdJoPaTo, line 278: Maybe it should be clarified that this is about the custom keyboards and not the inline keyboards?

site/docs/plugins/keyboard.md
```diff
@@ -273,6 +273,17 @@ await ctx.reply(text, {
...
+Your bot cannot differentiate between ordinary text messages, and text messages that were sent by clicking a button.
+
+Moreover, buttons will always send exactly the message that's written on them.
+Telegram does not allow you to create buttons that display one text, but send another.
```
EdJoPaTo, line 282: Maybe it should be highlighted that exactly that is possible with the inline keyboard and that its part of why the inline keyboard is far more powerful.

EdJoPaTo (dismissed):

site/docs/plugins/keyboard.md
KnorpelSenf: We already have a section with the exact same title under the inline keyboard section.

EdJoPaTo: "all that keyboards" still can sound misleading.

KnorpelSenf: I see that point. I'm not sure how to do about this, I don't want to say “keyboards (not inline keyboards)” everywhere

KnorpelSenf: What do you think about b7a28ef7c4932faf74abb84d8c363180ccf434ee?

EdJoPaTo: I remember them being referred to as "custom keyboards" in the documentation which seems more declarative than just "keyboard". Looking into the documentation its not always the case but its definitely done in many cases.

KnorpelSenf: The Bot API types use `Keyboard`. They may also use `ReplyKeyboard`. Be it custom keyboard or reply keyboard, I would not want to introduce a _third_ term. If we rename the `Keyboard` class, I would be open to using a different term consistently everywhere.

site/docs/plugins/keyboard.md
```diff
@@ -1,6 +1,23 @@
...
+| [**Inline Keyboard**](#inline-keyboards) | a set of buttons that is displayed underneath a message inside the chat                                                           |
+| **Keyboard button**                      | a button in a keyboard, sends a text message with its label when pressed                                                          |
+| **Inline keyboard button**               | a button in an inline keyboard, sends an callback query not visible to the user when pressed, sometimes just called inline button |
+| **`Keyboard`**                           | class in grammY to create keyboards                                                                                               |
+| **`InlineKeyboard`**                     | class in grammY to create inline keyboards                                                                                        |
+
+> Note that both keyboard buttons and inline keyboard buttons can also have other functions, such as requesting the user location, opening a website, and so on.
```
rojserbest, line 19:
```suggestion
> Note that both keyboard buttons and inline keyboard buttons can also have other functions, such as requesting the user's location, opening a website, and so on.
```

site/docs/plugins/keyboard.md
```diff
@@ -273,6 +290,18 @@ await ctx.reply(text, {
...
+### Responding to Clicks
+
+As mentioned earlier, all that keyboards do is sending regular text messages.
+Your bot cannot differentiate between ordinary text messages, and text messages that were sent by clicking a button.
+
+Moreover, buttons will always send exactly the message that's written on them.
+Telegram does not allow you to create buttons that display one text, but send another.
+If you need to do this, you should use an inline keyboard instead.
+
+In order to handle the click of a specific button, you can use `bot.hears` with the same text as you put on the button
```
rojserbest, line 302:
```suggestion
In order to handle the click of a specific button, you can use `bot.hears` with the same text as you put on the button.
```

rojserbest (changes requested):

rojserbest (approved):

site/docs/plugins/keyboard.md
EdJoPaTo: The type was Keyboard from the beginning without the knowledge of another appearing later. I dont think the types itself should be used as a single reference of what its called. As it is referred to as custom keyboard I would do that too as it clarifies a lot what is meant by "keyboard". This doesnt need to change the typings as they are mirroring the Telegram types but can change how they are referred in the docs. The goal of the grammY docs is to be easy to understand and a "keyboard" can differ with its meaning.

But just because I would approach it that way doesnt mean its the only/best way to go.

KnorpelSenf: I don't like about the term custom keyboards that it makes it sound like inline keyboard could not be custom-built. I think that can be confusing. Do you agree?

I don't like about the term reply keyboard that we already overload the word reply thrice (`reply_to_message_id`, `ctx.reply`, `reply_markup`) and I am not comfortable adding a fourth meaning. What do you think?

Overall, I am more comfortable with calling it custom keyboards and then casually describing inline keyboards as custom-built somewhere, rather than calling keyboards from now on reply keyboards.

EdJoPaTo: I don't think "Reply Keyboard" would be a good term for it. `sendMessage()` can also send that keyboard without being a reply. Which is also true for `reply_markup`. (Thinking about it… Telegraf named the class Markup and not ReplyMarkup which seems to make more sense. It might also because that name is shorter but still.)

About the naming and how I interpreted it so far:
You have your normal phone keyboard. It can be replaced with a custom keyboard typing predefined, custom texts instead of the letters the normal phone keyboard provides.
Also there is another keyboard attached to messages, with is in between the messages/lines → inline.

> it makes it sound like inline keyboard could not be custom-built

Hm… I guess I understand the thought behind it but doesnt seem very confusing to me. (Which doesnt mean it is a less valid argument.) Personally I think the inline_query and inline_keyboard stuff is more confusing. While the inline_keyboard is the the only thing possible for inline query results the inline keyboard also works independent of the inline query / inline mode.
I dont think we will get rid of every confusing aspect of the API but as long as we can describe everything by their distinct names we can explain things.

I think calling something A and something else BA while aspects of A dont work for BA and the other way around is more confusing than calling something with distinct CA and BA names. (This is the same problem with inheritance of classes which easily become inconsistent)
In the case of "custom keyboard" we wouldnt invent a new term as the official Bot API docs already call it that way.

KnorpelSenf: Alright. Then let's switch to calling them custom keyboards.

This requires the entire documentation to be updated, which includes
- this article
- all references to this article
- all references to header links (the titles will change)
- all TSDoc in the grammY core package


I can do this, but it may take some time.

EdJoPaTo: Not something that has to be updated all at the same time. Also something that can be done by someone willing to improve docs. Maybe create a separate Issue for that so others can see it being open and provide PRs more easily?

KnorpelSenf: I prefer very much to do all changes immediately and consistently. Nothing is worse than outdated docs. Please see https://github.com/grammyjs/grammY/pull/135

site/docs/plugins/keyboard.md
```diff
@@ -1,6 +1,23 @@
-# Keyboards and Inline Keyboards (built-in)
+# Inline and Custom Keyboards (built-in)
```
EdJoPaTo, line 1: "Inline and Custom" or "Custom and Inline"? The order differs

site/docs/plugins/keyboard.md
```diff
@@ -134,19 +151,23 @@ It makes sense to define `bot.on('callback_query:data')` at last to always answe
 Otherwise, some clients may display a loading animation for up to a minute when a user presses a button that your bot does not want to react to.
 :::
 
-## Keyboards
+## Custom Keyboards
 
-> Revisit the keyboard section in the [Introduction for Developers](https://core.telegram.org/bots#keyboards) written by the Telegram team.
+First things first: custom keyboards are sometimes just called keyboards, sometimes they're called reply keyboards, and even Telegram's own documentation is not consistent in this repect.
+As a simple rule of thumb, if it isn't absolutely obvious from the context is meant, every keyboard that is not an inline keyboard is a custom keyboard.
```
EdJoPaTo, line 157:
```suggestion
As a simple rule of thumb, when it isn't absolutely obvious from the context and not called inline keyboard, it probably is a custom keyboard.
```

site/docs/plugins/keyboard.md
```diff
@@ -260,26 +281,39 @@ await ctx.reply(text, {
...
+#### Selectively Send Custom Keyboards
 
-You can specify the `selective` option if you want to show the keyboard to only those users that are @-mentioned in the text of the message object, and to the sender of the original message in case your message is a reply.
+You can specify the `selective` option if you want to show the custom keyboard only to those users that are @-mentioned in the text of the message object, and to the sender of the original message in case your message is a [reply](/guide/basics.html#sending-messages-with-reply).
 
 ```ts
 await ctx.reply(text, {
   reply_markup: {
     selective: true,
+    reply_to_message_id: ctx.msg.message_id,
```
EdJoPaTo, line 292: `reply_to_message_id` is on the main object, not inside `reply_markup`.

EdJoPaTo (changes requested):

site/docs/plugins/keyboard.md
KnorpelSenf: This is intentional. It is now consistent with the article.

site/docs/plugins/keyboard.md
KnorpelSenf: Good point, thanks.

site/docs/plugins/keyboard.md
EdJoPaTo: Maybe to be more precise: This heading has a different order than the text in the vuepress config.ts

KnorpelSenf: Thank you, fixed.

EdJoPaTo (approved):
Perfectionists at work… Guess we have over-~~engineered~~documented(?) the keyboard part now :sparkles:

WingLim (approved):
🚀

PR #183, opened 2021-12-22
Do not repeat built-in

KnorpelSenf:
Follow-up for #175


roj1512 (approved):
Thanks for the good catch!

PR #196, opened 2021-12-29
Fix SSR mismatch

KnorpelSenf:
Attempt to fix https://github.com/vuepress/vuepress-next/discussions/551


KnorpelSenf:
Needs to be fixed by https://github.com/vuepress/vuepress-next/issues/602

PR #197, opened 2022-01-03
Fix Telegram chat link

KnorpelSenf:
Reported in https://t.me/grammyjs/32664


KnorpelSenf:
@habemuscode needs to be integrated in #180 if I'm not mistaken

roj1512 (approved):

PR #201, opened 2022-01-07
Document the emoji plugin

AndreoliBR:
Title says it all.

### DIFF
--- a/site/docs/plugins/emoji.md
+++ b/site/docs/plugins/emoji.md
@@ -82,7 +82,7 @@ bot.command("start", async (ctx) => {
 });
 ```
 
-:::warn Notice two things
+:::warn Keep in mind that
 `ctx.emoji` and `ctx.replyWithEmoji` **ALWAYS** use template strings. If you're unfamiliar with this syntax, you can read more about it [here](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Template_literals).
 :::
 
--- a/site/docs/plugins/emoji.md
+++ b/site/docs/plugins/emoji.md
@@ -82,14 +82,12 @@ bot.command("start", async (ctx) => {
 });
 ```
 
-::: warn Notice two things:
-
-1. `ctx.emoji` and `ctx.replyWithEmoji` **ALWAYS** use template strings.
-2. The emojis you want goes inside `${"templates"}`.
-   :::
+:::warn Notice two things
+`ctx.emoji` and `ctx.replyWithEmoji` **ALWAYS** use template strings. If you're unfamiliar with this syntax, you can read more about it [here](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Template_literals).
+:::
 
 ## Plugin Summary
 
 - Name: `emoji`
 - Source: <https://github.com/grammyjs/emoji>
-- Reference: TBD
+- Reference: https://doc.deno.land/https://deno.land/x/grammy_emoji/mod.ts
--- a/site/docs/plugins/emoji.md
+++ b/site/docs/plugins/emoji.md
@@ -30,8 +30,11 @@ You can install this plugin on your bot like this:
 ```ts
 import { EmojiFlavor, emojiParser } from "@grammyjs/emoji";
 
-type MyContext = Context & // You can pass your own context here
-  EmojiFlavor;
+// This is called Context Flavoring
+// You can read more about at:
+// https://grammy.dev/guide/context.html#context-flavors
+type MyContext = Context & EmojiFlavor;
+
 const bot = new Bot<MyContext>(""); // <-- put your bot token between the ""
 
 bot.use(emojiParser());
@@ -57,8 +60,11 @@ import {
   emojiParser,
 } from "https://deno.land/x/grammy_emoji/mod.ts";
 
-type MyContext = Context & // You can pass your own context here
-  EmojiFlavor;
+// This is called Context Flavoring
+// You can read more about at:
+// https://grammy.dev/guide/context.html#context-flavors
+type MyContext = Context & EmojiFlavor;
+
 const bot = new Bot<MyContext>(""); // <-- put your bot token between the ""
 
 bot.use(emojiParser());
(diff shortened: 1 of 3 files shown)


roj1512:
@dcdunkan you messed up with the formatting again. Please run `deno fmt --config deno.json` in the `site` directory.

roj1512:
@dcdunkan thanks!

site/docs/plugins/emoji.md
```diff
@@ -1 +1,68 @@
-Coming soon, please come back later
+# Emoji Plugin (`emoji`)
+
+With this plugin, you can easily insert emojis on your replies searching for them and inserting text instead of manually copying an emoji from web or Telegram and pasting at your code.
```
KnorpelSenf, line 3: I'd extend this part. It should first say that
- people use emoji in their code
- why that's bad/annoying (hard to edit on desktop, need to copy & paste from internet, not all desktop systems can render all emojis)
- how this plugin solves it (externalise emoji data, reference by name)
- uses template strings to have a super concise syntax (link to MDN, it's not a well-known features in JS)
- autocomplete is available!

Some of these points are already touched in the current version, but they're mixed in one long and hard to read sentence.

You may want to decide to only elaborate on a few of these points, then show the code examples early, and then explain the rest.

site/docs/plugins/emoji.md
```diff
@@ -1 +1,68 @@
...
+
+You can install this plugin on your bot like this:
+
+<CodeGroup>
+  <CodeGroupItem title="TypeScript" active>
+
+```ts
+import { EmojiFlavor, emojiParser } from "@grammyjs/emoji";
+
+type FlavoredContext =
```
KnorpelSenf, line 15: Maybe call it `MyContext` like all other plugins? Optionally, provide a relative link to the context flavouring if you want.

site/docs/plugins/emoji.md
```diff
@@ -1 +1,68 @@
...
+```
+
+</CodeGroupItem>
+</CodeGroup>
+
+And you can get emojis in your replies like this:
+
+```js
+bot.command("start", async (ctx) => {
+  await ctx.reply(ctx.emoji`Welcome! ${"smiling_face_with_sunglasses"}`);
```
KnorpelSenf, line 55: I liked the README comments that show how this will be rendered. Include them here, too?

site/docs/plugins/emoji.md
```diff
@@ -1 +1,68 @@
...
+});
+```
+
+Notice two things:
+
+1. `ctx.emoji` and `ctx.replyWithEmoji` **ALWAYS** use template strings.
```
KnorpelSenf, line 61: Could be formatted as a yellow warning block if you want. Format:
::: warn Title
Text
:::

KnorpelSenf (changes requested):

site/docs/plugins/emoji.md
AndreoliBR: I've used this syntax but it doesn't reflected on website. I already used `deno fmt emoji.md --config deno.json`.

Maybe @roj1512 could help me with this?

AndreoliBR:
Requesting help to @roj1512 

Warn block isn't working, I've already deno-formatted it.

KnorpelSenf (approved):
Thanks! This is good.

If you feel like making it excellent, you may want to add an example with `ctx.replyWithEmoji` as well as for sending emoji in other places, such as captions.

If you look at the API reference, it could also use some explanations, but that's nothing to do in this PR. It needs to be added to the source code.

dcdunkan:
What about adding an example for [this](https://t.me/grammyjs/32725)?
```ts
bot.hears(getEmoji("grinning_face"), ...)
```

KnorpelSenf:
I like the idea, but I'd change the code a bit because it's very uncommon for users to send single emoji to bots. How about this?

```ts
const reactions = [getEmoji("thumbs_up"), getEmoji ("thumbs_down")]
bot.callbackQuery(reactions, ctx => {})
```

dcdunkan:
Also, I think it is better to have links to Template Literals docs in `en-US` instead of `pt-BR`.
[L21](https://github.com/grammyjs/website/blob/00ec2a2bbafda9cab4c9b0596552741b7e60dad2/site/docs/plugins/emoji.md?plain=1#L21) and [L86](https://github.com/grammyjs/website/blob/00ec2a2bbafda9cab4c9b0596552741b7e60dad2/site/docs/plugins/emoji.md?plain=1#L86)

AndreoliBR:
> Also, I think it is better to have links to Template Literals docs in `en-US` instead of `pt-BR`. [L21](https://github.com/grammyjs/website/blob/00ec2a2bbafda9cab4c9b0596552741b7e60dad2/site/docs/plugins/emoji.md?plain=1#L21) and [L86](https://github.com/grammyjs/website/blob/00ec2a2bbafda9cab4c9b0596552741b7e60dad2/site/docs/plugins/emoji.md?plain=1#L86)

At some places it's already at en-US, I didn't pay attention to that! Gonna commit all fixes then it's ready to merge.

After these fixes I'll reserve some time to write better and full docs.

KnorpelSenf:
Sounds great!

Once you're done, you can add the label _ready for translation_. From then on, we should not change the files here anymore.

roj1512:
@AndreoliBR are you done?

roj1512 (approved):

AndreoliBR:
Fixes are done but later on I'll still add more things (this can delay translations) as KnorpelSenf said:
> If you feel like making it excellent, you may want to add an example with ctx.replyWithEmoji as well as for sending emoji in other places, such as captions.

I'll leave merging decision with you.

KnorpelSenf:
Those changes should be made in a follow-up PR. The plugin is already released, so it's good to get some docs out there. We can reiterate on them later.

@WingLim we won't touch this anymore, you're free to go :)

WingLim:
It seems this plugin not add to the header nav bar in `.vuepress/config.ts`?

WingLim (approved):
In English version:
Just add a little fix in `Plugin Summary - Reference`, everything else look good to me.

roj1512 (approved):

PR #222, opened 2022-01-28
Make quotes consistent

roj1512:
- Replaced the missed `"`s to `“` and `”`.

### DIFF
--- a/site/docs/README.md
+++ b/site/docs/README.md
@@ -43,7 +43,7 @@ import { Bot } from "grammy";
 
 const bot = new Bot(""); // <-- put your bot token here (https://t.me/BotFather)
 
-// Reply to any message with "Hi there!".
+// Reply to any message with “Hi there!”.
 bot.on("message", (ctx) => ctx.reply("Hi there!"));
 
 bot.start();
--- a/site/docs/README.md
+++ b/site/docs/README.md
@@ -57,7 +57,7 @@ const { Bot } = require("grammy");
 
 const bot = new Bot(""); // <-- put your bot token here (https://t.me/BotFather)
 
-// Reply to any message with "Hi there!".
+// Reply to any message with “Hi there!”.
 bot.on("message", (ctx) => ctx.reply("Hi there!"));
 
 bot.start();
@@ -71,7 +71,7 @@ import { Bot } from "https://deno.land/x/grammy/mod.ts";
 
 const bot = new Bot(""); // <-- put your bot token here (https://t.me/BotFather)
 
-// Reply to any message with "Hi there!".
+// Reply to any message with “Hi there!”.
 bot.on("message", (ctx) => ctx.reply("Hi there!"));
 
 bot.start();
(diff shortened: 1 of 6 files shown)


site/docs/README.md
```diff
@@ -43,7 +43,7 @@ import { Bot } from "grammy";
 
 const bot = new Bot(""); // <-- put your bot token here (https://t.me/BotFather)
 
-// Reply to any message with "Hi there!".
+// Reply to any message with “Hi there!”.
```
KnorpelSenf, line 46: I'm not sure we should use non-ASCII256 characters in code. Maybe I have pre-JS mentality here, did we already do this in any other places before?

I do realise that the Chinese translation has Chinese characters in the comments, and I believe that this is the right thing to do—what's your opinion on the matter?

roj1512: I've already changed the comments too. I forgot this one.

roj1512: Let me fix the comments too.

KnorpelSenf (approved):

site/docs/guide/files.md
```diff
@@ -159,7 +159,7 @@ On Deno, you can send `Blob` objects, too.
 ```ts
 // Send a buffer or a byte array.
 const buffer = Uint8Array.from([65, 66, 67]);
-new InputFile(buffer); // "ABC"
+new InputFile(buffer); // “ABC”
```
KnorpelSenf, line 162: These things are debatable. Using `"ABC"` would refer to the JavaScript value, using “ABC” would refer to the content as understood by humans. I'm okay with both ways, I just want to make sure that this was considered and a deliberate choice.

roj1512: I haven't touched the quotes referring to JavaScript values. For example in the quotes in [this line](https://github.com/grammyjs/website/blob/e89ca6901af89c09b921331fa2d62d6b7d1906a1/site/docs/guide/getting-started.md?plain=1#L57).

KnorpelSenf: I saw—not talking about that.

I do mean the comments. The old version was saying, _this buffer is equivalent to putting `"ABC"` into a file_, and the new version is saying, _if you look at the file it will contain the word “ABC”_. This is implied by using different quotes. Does this make more sense?

roj1512: I'll fix it.

KnorpelSenf: I'm not saying this is an isse, there are two options and I am fine with both. My point was rather that we both agree that these two options exist, and that we make an active choice, rather than performing these changes accidentally, since they are changing the meaning :)

KnorpelSenf: So just go ahead with whatever you prefer, be it changing things back or leaving them as-is.

roj1512: We'll use `"` if the contents inside the quotes are *only* referring to data. How's this?

KnorpelSenf: Sounds good! How do you manage to remember all these things? :smile:

PR #230, opened 2022-02-01
Uz/guide/getting started

Dostonlv:


site/docs/.vuepress/config.ts
```diff
@@ -700,7 +1032,50 @@ export default defineUserConfig<DefaultThemeOptions>({
...
+                noResultsScreen: {
+                  noResultsText: "So'rov bo'yicha natija mavjud emas",
+                  suggestedQueryText: "Quyidagini qidirishga urinib ko'ring",
+                  openIssueText:
+                    "Bu so'rovdan natija chiqishiga ishonchingiz komilmi?",
+                  openIssueLinkText: "Bizga xabar bering",
+                },
+              },
+            },
+          },
```
KnorpelSenf, line 1077: @WingLim @linbuxiao @taotie111 @JiquanWang99 could we also have these translations for Chinese (in a second PR)? The concern the Algolia search interface. We missed this in the early days when we were translating the config.

KnorpelSenf: cc @kidonng

KnorpelSenf: @IamItachiUchiha @mcpeblocker @sad1go0 could you leave a link here which points to the English originals of this configuration that you used when translating this? Such as the Algolia API reference or something.

mcpeblocker: Here:
https://docsearch.algolia.com/docs/api/#translations

mcpeblocker:
Sorry for that PR :(

PR #241, opened 2022-02-14
`bot.errorBoundary`-free error boundary example

roj1512:
[grammyjs/39335](https://t.me/grammyjs/39335)-[39360](https://t.me/grammyjs/39360)

### DIFF
--- a/site/docs/guide/errors.md
+++ b/site/docs/guide/errors.md
@@ -100,7 +100,7 @@ If you divide your code base into different parts, _error boundaries_ allow you 
 They achieve this by letting you fence errors in a part of your middleware.
 In other words, if an error is thrown in a specially protected part of middleware, it will not be able to escape from that part of the middleware system.
 Instead, a dedicated error handler is invoked, and the surrounded part of the middleware pretends to complete successfully.
-This is a feature of grammY’s middleware system, so error boundaries don’t care whether you’re running your bot with webhooks or long polling.
+This is a feature of grammY's middleware system, so error boundaries don't care whether you're running your bot with webhooks or long polling.
 
 Optionally, you may choose to instead let the middleware execution _resume_ normally after the error was handled, continuing right outside the error boundary.
 In that case, the fenced middleware does not only act as if it had completed successfully, but it also passes on the control flow to the next middleware that was installed after the error boundary.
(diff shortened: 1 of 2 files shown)


site/docs/guide/errors.md
```diff
@@ -100,7 +100,7 @@ If you divide your code base into different parts, _error boundaries_ allow you
 They achieve this by letting you fence errors in a part of your middleware.
 In other words, if an error is thrown in a specially protected part of middleware, it will not be able to escape from that part of the middleware system.
 Instead, a dedicated error handler is invoked, and the surrounded part of the middleware pretends to complete successfully.
-This is a feature of grammY's middleware system, so error boundaries don't care whether you're running your bot with webhooks or long polling.
+This is a feature of grammY’s middleware system, so error boundaries don’t care whether you’re running your bot with webhooks or long polling.
```
KnorpelSenf, line 103: Should this be applied consistently to all articles? I do understand if you don't think that this would be worth the effort.

site/docs/guide/errors.md
```diff
@@ -137,13 +137,33 @@ function errorHandler(err: BotError) {
...
+const composer = new Composer();
+
+const protected = composer.errorBoundary(boundaryHandler);
+protected.use(/* B */);
+
+bot.use(composer);
+bot.use(/* C */);
+
+function boundaryHandler(err: BotError, next: NextFunction) {
+  throw new Error("Error in B!");
```
KnorpelSenf, line 159: This needs more explanation, similar to the comment in the sniper above.

KnorpelSenf (changes requested):

site/docs/guide/errors.md
KnorpelSenf: Do you want to perform these changes for all articles in a separate PR then?

site/docs/guide/errors.md
KnorpelSenf: I was rather thinking for something like this:
```suggestion
  console.error("Error in Q, X, Y, or Z!", err);
  /*
   * You could call `next` if you want to run
   * the middleware at C in case of an error:
   */
  // await next()
  throw new Error("Error in B!");
```

roj1512:
@KnorpelSenf please review now.

WingLim (approved):
🚀

PR #242, opened 2022-02-15
Fix typo

KnorpelSenf:
May be integrated into #241


KnorpelSenf:
Feel free to change the target branch

roj1512 (approved):
Won’t integrate to #241 because “not having typos” has a higher priority than having more examples.

PR #243, opened 2022-02-15
Initialize the Notifier

roj1512:


roj1512:
Please let me know if we need to have any text talking about the bot and why is it existing.

notifier/mod.ts
```diff
@@ -0,0 +1,47 @@
+import { Bot } from "https://deno.land/x/grammy/mod.ts";
+import { Application } from "https://deno.land/x/oak/mod.ts";
```
KnorpelSenf, line 2: No versions?

notifier/env.ts
```diff
@@ -0,0 +1,8 @@
+import "https://deno.land/x/dotenv/load.ts";
+import { cleanEnv, num, str } from "https://deno.land/x/envalid/mod.ts";
```
KnorpelSenf, line 2: No versions?

KnorpelSenf (changes requested):
I would add a short README file, just 3 lines or so, and a link to the translation guide. Don't spend more than 5 min on this.

roj1512:
We can also talk about the private translators group and this bot in the Translation Guide, but I personally think it is not needed because saying that translators will have “push access” does it all.

KnorpelSenf (approved):
Thanks!

PR #249, opened 2022-02-20
Run deploy and format workflows only for `.md`

roj1512:


KnorpelSenf (commented):
Shouldn't it maybe format the bot code as well?

roj1512:
Sure, but we don't want to deploy what was pushed to the bot files. Maybe we will separate the deployment to another file?

KnorpelSenf:
Yes that makes sense.

Instead, could we always run the formatting but only deploy for markdown changes? I think it's good to only deploy if the formatting is correct.

roj1512:
The current state of the pull request:

- Uses one Deno config to format all files supported by the Deno Formatter.
- Builds the website if a `.md` file was modified, regardless of the format workfow passing.

We can also do this:

- Use two Deno configs, one in the root which ignores the `site/` directory, and the other in the `site/` directory and dedicated to it.
- Build the website only if a `.md` in the `site/` directory was modifed, and only if the format of the directory passed.

@KnorpelSenf, please let me know if we should continue in the current state or do my other possible suggestion.

KnorpelSenf:
I see! The current state is fine, I think I didn't read thoroughly enough. Go ahead 👍🏻

KnorpelSenf (approved):

PR #252, opened 2022-02-23
Document the Fluent plugin

roj1512:


WingLim (approved):
🚀

KnorpelSenf (approved):
LGTM!

In the long run, we may consider to deprecate the i18n plugin. I am not sure about the chances of receiving a 2.0 release. Some community work has been going into this for almost a year now, mostly by @HeeroML and @EdJoPaTo. However, such a release does not seem imminent. Having two official plugins that solve the same problem in different ways is not optimal.

KnorpelSenf (approved):
LGTM!

In the long run, we may consider to deprecate the i18n plugin. I am not sure about the chances of receiving a 2.0 release. Some community work has been going into this for almost a year now, mostly by @HeeroML and @EdJoPaTo. However, such a release does not seem imminent. Having two official plugins that solve the same problem in different ways is not optimal.

HeeroML:
I am writing you on telegram. Though replacing it isn't that bad of an idea, I would still try to finish 2.0

PR #264, opened 2022-03-09
[Feat] [Translations] [Spanish]: resources/faq

habemuscode:


roj1512 (commented):
**Outdated**

- Missing description for the `node-fetch` error.

**Recommendation**

The status texts should not be translated. For example, 404 Not Found and similar should be kept as they are.

habemuscode:
@roj1512 

All file reviewed since `node-fetch` is not the only change

roj1512 (approved):

PR #266, opened 2022-03-27
Do not specify --config in README

KnorpelSenf:
Deno can now detect config files automatically, so there's no need to `--config deno.json` when formatting the source code.


roj1512 (approved):
Thanks!

PR #267, opened 2022-03-28
Add proxy guide for Deno

MrSaeedNasiri:


site/docs/advanced/proxy.md
```diff
@@ -9,6 +9,26 @@ grammY let's you configure a number of things about how network requests are per
 This includes injecting a custom payload into every request, which can be used to install a proxy agent.
 Check out the `ApiClientOptions` in the [grammY API Reference](https://doc.deno.land/https://deno.land/x/grammy/mod.ts/~/ApiClientOptions).
 
+In Deno, here is how you would use a `http` proxy:
```
KnorpelSenf, line 12:
```suggestion
In Deno, here is how you would use an `http` proxy:
```

site/docs/advanced/proxy.md
```diff
@@ -9,6 +9,26 @@ grammY let's you configure a number of things about how network requests are per
...
+```ts
+import { Bot } from "https://deno.land/x/grammy/mod.ts";
+
+const client = Deno.createHttpClient({
+  proxy: { url: "http://host:port/" },
+});
+const bot = new Bot(TOKEN, {
+  client: {
+    baseFetchConfig: {
+      // @ts-ignore
```
KnorpelSenf, line 23:
```suggestion
```
This compiles for me without the `@ts-ignore` directive. Can you confirm?

site/docs/advanced/proxy.md
```diff
@@ -9,6 +9,26 @@ grammY let's you configure a number of things about how network requests are per
...
+      client,
+    },
+  },
+});
+```
+
+> Note that run with the '--unstable' flag
```
KnorpelSenf, line 30:
```suggestion
> Note that you need to run this with the `--unstable` flag.
```

KnorpelSenf (changes requested):
Thanks for the contribution!

I have a few minor comments:

site/docs/advanced/proxy.md
MrSaeedNasiri: no its no compile.
deno 1.20.3

```
error: TS2322 [ERROR]: Type '{ client: Deno.HttpClient; }' is not assignable to type 'Omit<RequestInit, "method" | "headers" | "body">'.
  Object literal may only specify known properties, and 'client' does not exist in type 'Omit<RequestInit, "method" | "headers" | "body">'.
```

KnorpelSenf: I see, I can reproduce it now. This is because the API is still unstable. grammY will extract the options object from the type signature of `fetch`, but it looks like this is not possible for different overloads.

Your example is correct, let's leave it as-is. Once the API is stabilised and the types are updated, we can remove the `@ts-ignore` directive.

KnorpelSenf (approved):
Great work!

roj1512 (approved):

roj1512:
@KnorpelSenf sorry for the latency, I missed this.

WingLim (approved):
🚀

PR #271, opened 2022-04-08
Update throttler docs to use runner to strongly suggest the use of a concurrent update handling mechanism

KnightNiwrem:


site/docs/plugins/transformer-throttler.md
```diff
@@ -30,14 +31,17 @@ const throttler = apiThrottler();
 bot.api.config.use(throttler);
 
 bot.command("/example", (ctx) => ctx.reply("I am throttled"));
-bot.start();
+
+// If you are using throttler, you most likely want to use a runner to handle update concurrently
```
KnorpelSenf, line 35:
```suggestion
// If you are using throttler, you most likely want to use a runner to handle updates concurrently
```

site/docs/plugins/transformer-throttler.md
```diff
@@ -50,14 +54,17 @@ const throttler = apiThrottler();
...
```
KnorpelSenf, line 58:
```suggestion
// If you are using throttler, you most likely want to use a runner to handle updates concurrently
```

site/docs/plugins/transformer-throttler.md
```diff
@@ -70,7 +77,9 @@ const throttler = apiThrottler();
...
```
KnorpelSenf, line 81:
```suggestion
// If you are using throttler, you most likely want to use a runner to handle updates concurrently
```

KnorpelSenf (changes requested):
Thanks!

Tiny typo:

KnorpelSenf (approved):
LGTM

KnorpelSenf:
This wasn't supposed to be merged.

I have a suspicion that the standard comment format is different and that things should be aligned for consistency, but @roj1512 knows best about this.

We also need to update translations after @roj1512 has reviewed.

All follow-up changes can be performed in a new PQ, I restored the branch.

KnightNiwrem:
> This wasn't supposed to be merged.
> 
> I have a suspicion that the standard comment format is different and that things should be aligned for consistency, but @roj1512 knows best about this.
> 
> We also need to update translations after @roj1512 has reviewed.
> 
> All follow-up changes can be performed in a new PQ, I restored the branch.

Whoops. Is there any way to implement some kind of PQ checklist mechanism to make it clear that the PQ handling is different here?

KnorpelSenf:
There's not a checklist of tasks to do, but rather a distribution of responsibilities between several people. After a contribution, they take over with their well-established process. You can read more about this in the contribution guide, but the TL;DR is: perform the changes to the content, the rest will be taken care of.

KnorpelSenf:
@roj1512 @WingLim can you take care of this?

PR #272, opened 2022-04-09
Talk about a TypeScript build error in the FAQ

roj1512:


roj1512:
@KnorpelSenf Please give me feedback on the content, and tell me if it is in the right section with the right title.

site/docs/resources/faq.md
```diff
@@ -115,6 +115,12 @@ There are a few things you can do.
 4. Come ask us in the group chat for help. We have experienced people there.
 5. It is possible to ask Telegram to increase the limits, but this is very unlikely to happen if you did not do steps 1-3 first.
 
+### Cannot find type definition file for 'node-fetch'.
+
+This is the result of a version conflict between the installed `@types/node-fetch`, and the grammY dependency `node-fetch`.
```
KnorpelSenf, line 120:
```suggestion
This is the result of a version conflict between the installed `@types/node-fetch`, and the grammY dependency `node-fetch`.
```

KnorpelSenf, line 120: It's actually just missing type information. There's no mismatch. We don't have the types at a production dependency, so they're missing if users try to type-check their node modules.

Skipping this step of checking everything is the best solution.

Alternatively, users can install the package `@types/node-fetch@2`.

roj1512: I modified it.

site/docs/resources/faq.md
```diff
@@ -115,6 +115,12 @@ There are a few things you can do.
 4. Come ask us in the group chat for help. We have experienced people there.
 5. It is possible to ask Telegram to increase the limits, but this is very unlikely to happen if you did not do steps 1-3 first.
 
+### Cannot find type definition file for 'node-fetch'.
+
+This is the result of some missing type declarations.
+
+The recommended way to fix this is to set `skipLibCheck` to `true` in your TypeScript compile options.
+
```
KnorpelSenf, line 123:
```suggestion

If you are sure that you need this option to be kept to `false`, you can instead install the missing type definitions by running `npm i -D @types/node-fetch@2`.
```

roj1512:
@KnorpelSenf for some reason the button to request another review is not working.

KnorpelSenf (approved):

site/docs/resources/faq.md
```diff
@@ -115,6 +115,14 @@ There are a few things you can do.
 4. Come ask us in the group chat for help. We have experienced people there.
 5. It is possible to ask Telegram to increase the limits, but this is very unlikely to happen if you did not do steps 1-3 first.
 
+### Cannot find type definition file for 'node-fetch'.
```
WingLim, line 118: According to https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md026

We should remove this punctuation.

WingLim (changes requested):

site/docs/resources/faq.md
KnorpelSenf: I agree.
```suggestion
### Cannot find type definition file for 'node-fetch'
```

WingLim: Or this is the orignal error message which we should keep it what it is?

roj1512: It was the original message, but I think you can remove the dot. The meaning will stay there.

KnorpelSenf (approved):

WingLim (approved):

PR #278, opened 2022-04-12
Automate smart punctuation

roj1512:
- Reverts #222.
- Replaces “” with ".
- Makes all quotes converted to smart ones automatically leaving the code blocks as they are.


CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ We have observed that the following workflow works well, and we usually try to s
 
 1. You (or someone else) starts a PR of unknown quality with a change to the English docs.
 2. Someone else (or you) reviews and comments it, the participants iterate over the changes until they're good.
-3. A member of @grammyjs approves the PR and adds the label “read for translation”.
+3. A member of @grammyjs approves the PR and adds the label "read for translation".
```
KnorpelSenf, line 20: I think this file won't be transformed, so we could leave “”.

TRANSLATING.md
```diff
@@ -91,16 +91,16 @@ At this point, all you need to do is to keep it that way.
 Whenever an English article is updated (by us, you, or anyone else), a pull request will be opened.
 It usually takes some time before we get the content right, so don't bother translating anything at this point yet.
 
-Once the content is finalised, we add the “ready for translation” label.
+Once the content is finalised, we add the "ready for translation" label.
```
KnorpelSenf, line 94: Ditto

KnorpelSenf (changes requested):
This is still missing a plugin to convert the quotation marks, isn't it? The preview is rendering ASCII256 characters.

roj1512:
@KnorpelSenf It doesn’t need a plugin. I had a typo, it is fixed now. The quotes should be fine in the next deployment.

KnorpelSenf (approved):
Pretty cool, I wasn't aware that VuePress could do this. Should've done this from the start!

roj1512:
Yes, pretty cool we finally came with this solution! We were about to solve a problem by making other problems.

PR #281, opened 2022-04-16
Add Glossary

KnorpelSenf:
There are a large number of confusing terms in the Bot API. They all sound similar, but they are very different.

We should have a central place where these terms are
- mentioned
- briefly defined
- linked with their source definition

and things are simply sorted alphabetically.

This PQ is incomplete, we need more definitions. Some ideas:

- reply keyboard
- custom keyboard
- custom reply keyboard
- keyboard button
- inline keyboard
- inline menu
- inline keyboard button
- callback query
- inline query
- switch_inline_query button
- etc

@roj1512 can you continue with this, or delegate it to someone who does?


roj1512:
@KnorpelSenf Do you gift the new commit a yes, or a no?

KnorpelSenf:
> @KnorpelSenf Do you gift the new commit a yes, or a no?

What do you mean?

roj1512:
@KnorpelSenf I mean how is this https://github.com/grammyjs/website/pull/281/commits/53d8529e719b48fb4ed4752e8e899c1ebafc5dc4

KnorpelSenf:
Oh I see. I find that this makes it harder to read. If I encounter the term reply keyboard, I would like to scan the leftmost column for it and find it there. The description can then be “Alias for custom keyboard” so I know where to look instead. This is better than having to scan two columns whenever I want to see the definition for a word (one of which contains a comma-separated list of terms).

KnorpelSenf:
> You mean aliases should have their own row, right?

Correct.

KnorpelSenf:
> > This PQ is incomplete, we need more definitions
> 
> Please consider adding "PQ" to the glossary.

I agree, we certainly need this. It would be fun to be able to throw a rude _RTFM_ at people who ask about that :smile:

KnorpelSenf:
There has been neither activity nor any more demand. Some confusing terms about keyboards are explained there. Some other confusing terms will be renamed in the library for the next major release. The remaining ones are not similar enough to justify this change. Closing.

PR #305, opened 2022-04-27
[Feat] [Translations] [Spanish]: guide/commands

roj1512:


roj1512 (commented):
**Oudated**

habemuscode:
@roj1512 I don't see the outdated part in the git blame

roj1512:
I must have mixed this with another page. This was last updated in November and your translations were around February. So this is fine, my bad :)

PR #321, opened 2022-05-04
Describe how filter query logic and types can be reused

KnorpelSenf:
As discussed in https://t.me/grammyjs/48027

### DIFF
--- a/site/docs/guide/filter-queries.md
+++ b/site/docs/guide/filter-queries.md
@@ -277,7 +277,7 @@ bot.on("message").filter(
 );
 ```
 
-### Reusing Filter Query Logic and Types
+### Reusing Filter Query Logic
 
 Internally, `bot.on` relies on a function called `matchFilter`.
 It takes a filter query and compiles it down to a predicate function.
@@ -291,7 +291,9 @@ For example, you can decide to drop all updates that match a certain query:
 bot.drop(matchFilter(":text"));
 ```
 
-Analogously, the same thing works on the type level.
+Analogously, you can make use of the filter query types that grammY uses internally:
+
+### Reusing Filter Query Types
 
 Internally, `matchFilter` uses TypeScript's [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to narrow down the type of `ctx`.
 It takes a type `C extends Context` and a `Q extends FilterQuery` and produces `ctx is Filter<C, Q>`.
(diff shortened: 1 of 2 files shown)


Satont:
Can `matchFilter ` and `Filter` be splitted as different sections? So It will be easier to link to each one. I saw so many questions in grammY chat about how to do `Filter`.

Satont:
LGTM 👍🏻

Satont (approved):

roj1512 (approved):

KnorpelSenf:
Why force push? What happened here?

roj1512:
Hmm, I didn't do anything around this, to be honest.

roj1512:
Perhaps it is a bug.

KnorpelSenf:
Oh I see, it was the target branch `main` which was force-pushed

roj1512:
Yes, I did it on `main` and `add-deno-bin`. They are going to be my last force pushes anyway.

KnorpelSenf:
I appreciate that :D

site/docs/guide/filter-queries.md
```diff
@@ -319,3 +354,6 @@ These functions are then combined to form a predicate that only has to check for
 
 This system uses less operations than some competing libraries, which need to perform containment checks in arrays when routing updates.
 grammY's filter query system is much more powerful.
+
+```
```
WingLim, line 358: Why we have this without any code in it?

site/docs/guide/filter-queries.md
```diff
@@ -277,6 +277,41 @@ bot.on("message").filter(
...
+
+You can import `matchFilter` directly if you want to use it in your own logic.
+For example, you can decide to drop all updates that match a certain query:
+
+```ts
+// Drop all text messages or text channel posts.
+bot.drop(matchFilter(":text"));
+```
+
+Analogously, you can make use of the filter query types that grammY uses internally:
```
WingLim, line 294:
```suggestion
Analogously, you can make use of the filter query types that grammY uses internally.
```
Should it be a `.` not `:` ?

site/docs/guide/filter-queries.md
roj1512: Fixed.

KnorpelSenf: That's an artefact of formatting Markdown with incomplete code environments. The two lines should be removed.

roj1512 (approved):

PR #324, opened 2022-05-05
Let people without Deno format the docs

roj1512:
This is done by adding the `deno-bin` npm package to the development dependencies. That way, everyone who runs `npm install` will also get a local Deno binary in their `node_modules` folder. This can in turn be used to format the docs without global Deno installation on the system.

Given the size of `node_modules`, this will hardly be noticed. Just TypeScript has almost twice the size of the Deno binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,5 +40,4 @@ You can also run
 npm run docs:fmt
 ```
 
-to perform the formatting.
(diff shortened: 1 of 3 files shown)


roj1512:
Was approved in #322.

PR #327, opened 2022-05-05
Let people without Deno format the docs

KnorpelSenf:
This is done by adding the `deno-bin` npm package to the development dependencies. That way, everyone who runs `npm install` will also get a local Deno binary in their `node_modules` folder. This can in turn be used to format the docs without global Deno installation on the system.

Given the size of `node_modules`, this will hardly be noticed. Just TypeScript has almost twice the size of the Deno binary.

This PQ also fixes the sitemap setup and updates vuepress.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,1 +1,14 @@
-{ "deno.config": "./deno.json", "deno.enable": true }
+{
+  "deno.config": "./deno.json",
+  "deno.enable": true,
+  "deno.enablePaths": [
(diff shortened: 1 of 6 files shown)


KnorpelSenf:
After this is merged, I will trigger a manual recrawl of the site. The added sitemap will finally add all missing URLs again to the index.

roj1512 (approved):
Thank you!

KnorpelSenf:
image
Much better search coverage now. This was so worth it.

PR #338, opened 2022-05-10
Update the Deno Deploy guide

Loskir:


roj1512 (changes requested):

roj1512 (changes requested):

site/docs/hosting/deno-deploy.md
```diff
@@ -49,15 +53,31 @@ serve(async (req) => {
...
+
+### Method 2: With `deployctl`
+
+> This is a method for more advanced users. It allows you to deploy the project via the command line or Github Actions.
+
+1. Visit your [Deno Deploy dashboard](https://dash.deno.com/projects).
+2. Click on "New Project", then "Empty Project".
+3. Install [`deployctl`](https://github.com/denoland/deployctl)
+4. It's recommended to store the token in environment variables. You can set them up on the Settings page.
+5. Set your bot's webhook URL to `https://<PROJECT_NAME>.deno.dev/<TOKEN>` (replacing `<...>` with respective values). In order to do that, run this URL (in your browser, for example): `https://api.telegram.org/bot<TOKEN>/setWebhook?url=https://<PROJECT_NAME>.deno.dev/<TOKEN>`
```
roj1512, line 70:
```suggestion
```

site/docs/hosting/deno-deploy.md
```diff
@@ -49,15 +53,31 @@ serve(async (req) => {
...
+6. Create a new [access token](https://dash.deno.com/user/access-tokens). Save it somewhere
+7. Run this command to deploy: `deployctl deploy --project <PROJECT_NAME> ./mod.ts --prod --token <ACCESS_TOKEN>`
+8. To set up Github Actions, refer to [this](https://github.com/denoland/deployctl/blob/main/action/README.md).
```
roj1512, line 73:
```suggestion
4. Create a new [access token](https://dash.deno.com/user/access-tokens). Save it somewhere
5. Run this command to deploy: `deployctl deploy --project <PROJECT_NAME> ./mod.ts --prod --token <ACCESS_TOKEN>`
6. To set up Github Actions, refer to [this](https://github.com/denoland/deployctl/blob/main/action/README.md).
```

site/docs/hosting/deno-deploy.md
```diff
@@ -49,15 +53,31 @@ serve(async (req) => {
...
-### Method 2: With URL
+### Method 3: With URL
 
 > All you need for following this method to deploy your grammY bot, is a public URL to your `mod.ts` file.
 
 1. Create a new project on Deno Deploy.
 2. Click "Deploy URL".
 3. Input the public URL to your `mod.ts` file, and click "Deploy".
+4. It's recommended to store the token in environment variables. You can set them up on the Settings page.
+5. Set your bot's webhook URL to `https://<PROJECT_NAME>.deno.dev/<TOKEN>` (replacing `<...>` with respective values). In order to do that, run this URL (in your browser, for example): `https://api.telegram.org/bot<TOKEN>/setWebhook?url=https://<PROJECT_NAME>.deno.dev/<TOKEN>`
```
roj1512, line 83:
```suggestion

### Note

After getting your app running, you should configure your bot's webhook settings to point to your app.
To do that, send a request to https://api.telegram.org/bot<token>/setWebhook?url=<url>, replacing `<token>` with your bot's token, and `<url>` with the full URL of your app along with the path to the webhook handler.
```

site/docs/hosting/deno-deploy.md
```diff
@@ -49,15 +53,31 @@ serve(async (req) => {
...
 3. Visit your [Deno Deploy dashboard](https://dash.deno.com/projects).
-4. Create a new project.
-5. Scroll to the "Deploy from GitHub" section and click "Continue".
-6. Install the GitHub app on your account or organization and choose your repository.
-7. Select the branch you want to deploy, and then choose your `mod.ts` file to be deployed.
+4. Click on "New Project" and go to the "Deploy from GitHub repository" section.
+5. Install the GitHub app on your account or organization and choose your repository.
+6. Select the branch you want to deploy, and then choose your `mod.ts` file to be deployed.
+7. It's recommended to store the token in environment variables. You can set them up on the Settings page.
+8. Set your bot's webhook URL to `https://<PROJECT_NAME>.deno.dev/<TOKEN>` (replacing `<...>` with respective values). In order to do that, run this URL (in your browser, for example): `https://api.telegram.org/bot<TOKEN>/setWebhook?url=https://<PROJECT_NAME>.deno.dev/<TOKEN>`
```
roj1512, line 60:
```suggestion
```

roj1512 (changes requested):

roj1512:
@Loskir You can accept multiple suggestions together on the Files tab: https://github.com/grammyjs/website/pull/338/files

site/docs/hosting/deno-deploy.md
```diff
@@ -49,15 +53,30 @@ serve(async (req) => {
...
 > All you need for following this method to deploy your grammY bot, is a public URL to your `mod.ts` file.
 
 1. Create a new project on Deno Deploy.
 2. Click "Deploy URL".
 3. Input the public URL to your `mod.ts` file, and click "Deploy".
+
+### Note
+
+After getting your app running, you should configure your bot's webhook settings to point to your app.
+To do that, send a request to https://api.telegram.org/bot<token>/setWebhook?url=<url>, replacing `<token>` with your bot's token, and `<url>` with the full URL of your app along with the path to the webhook handler.
```
KnorpelSenf, line 82:
```suggestion
To do that, send a request to <https://api.telegram.org/bot<token>/setWebhook?url=<url>>, replacing `<token>` with your bot's token, and `<url>` with the full URL of your app along with the path to the webhook handler.
```

site/docs/hosting/deno-deploy.md
```diff
@@ -49,15 +53,30 @@ serve(async (req) => {
...
 > All you need for following this method to deploy your grammY bot, is a public URL to your `mod.ts` file.
 
 1. Create a new project on Deno Deploy.
 2. Click "Deploy URL".
 3. Input the public URL to your `mod.ts` file, and click "Deploy".
+
+### Note
+
+After getting your app running, you should configure your bot's webhook settings to point to your app.
+To do that, send a request to <https://api.telegram.org/bot<token>/setWebhook?url=<url>>, replacing `<token>` with your bot's token, and `<url>` with the full URL of your app along with the path to the webhook handler.
```
KnorpelSenf, line 82:
```suggestion
To do that, send a request to `https://api.telegram.org/bot<token>/setWebhook?url=<url>`, replacing `<token>` with your bot's token, and `<url>` with the full URL of your app along with the path to the webhook handler.
```

site/docs/hosting/deno-deploy.md
```diff
@@ -15,27 +15,31 @@ It's just like the other many serverless platforms, but dedicated for Deno apps.
 
 ```ts
 import { webhookCallback } from "https://deno.land/x/grammy/mod.ts";
+import { serve } from "https://deno.land/x/sift@0.5.0/mod.ts";
```
KnorpelSenf, line 18: Should we specify a version always or never? Or maybe never for grammY, but for everything else?

Loskir: I think the ideal scenario would be to always automatically specify the latest version. But it might be tricky to implement

roj1512: @Loskir The grammY version won't be a problem if not specified, because they are in their own documentation. What do you think about the 3rd parties?

roj1512: It might also not be that tricky to implement the auto-upgrade of the grammY imports. It is just a custom plugin away.

KnorpelSenf: Yes, doing automatic replacements in the code is hard, it would either have to be fetched dynamically for each page visit, or we would have to change the docs for every release. If there was VuePress for Deno, we could render things easily on the edge, but this does not exist yet.

I think it's not a problem with our own URLs. All code examples will work, regardless of the version. Also, the Deno LSP automatically suggests to replace the URL with the redirect, so if the code was to be copied into an editor, the latest version can be inserted automatically.

I'm rather concerned about third-party code, as we have no control over it.

roj1512: @KnorpelSenf For 3rd-party modules, I think we can specify the version that worked for us at the time of the writing of the documentation, and we can accept working PQs that upgrade their versions. What do you say?

KnorpelSenf: That sounds good, then lets stick with:
1. Never versions for grammY so that people always automatically pull in the lastest code
2. Always versions for third-party modules so that the code examples are guaranteed to work

@Loskir objections?

Loskir: > Also, the Deno LSP automatically suggests to replace the URL with the redirect, so if the code was to be copied into an editor, the latest version can be inserted automatically.

That's why i suggested specifying it in the docs. I'd like the examples to be completely copy-pasteable.

As for third-party libraries: I agree.

KnorpelSenf: I think the easiest way to achieve this is by creating a new plugin in the `config.ts` file. I suggest leaving this PQ as-is, and taking care of that on a different branch.

The Deno manual does it via env var syntax: https://github.com/denoland/manual/blob/main/introduction.md#comparison-to-nodejs

We could simply use regular expressions. I am sure that something like `@GRAMMY_VERSION` does not occur anywhere else.

I opened https://github.com/grammyjs/website/issues/341.

roj1512 (approved):

habemuscode:
In all docs, we have an empty line at the final of the file

KnorpelSenf:
As long as CI passes, things are okay. Minor formatting differences that the tooling doesn't catch are too hard to pursue.

roj1512:
Is it a problem?

site/docs/hosting/deno-deploy.md
```diff
@@ -15,27 +15,31 @@ It's just like the other many serverless platforms, but dedicated for Deno apps.
...
-  return new Response();
+    return new Response();
+  },
+  "/": () => {
+    return new Response("Hello world!");
+  },
 });
 ```
 
+We advise you to have your handler on some secret path rather than root. Here, we are using the bot token.
```
WingLim, line 41: I recommand that we could add the example for it.

```suggestion
We advise you to have your handler on some secret path rather than root(`/`). Here, we are using the bot token(`/<bot token>`).
```

KnorpelSenf: I agree.

KnorpelSenf:
The Spanish article must be synced on the other branch, right?

roj1512 (approved):

PR #348, opened 2022-05-12
Increase deployment timeout to 5 minutes

KnorpelSenf:
This will mostly eliminate the problem of netlify deploys failing randomly. With three languages, the output folder is growing large enough that this becoming more and more of an issue.


roj1512 (approved):
Thanks!

PR #351, opened 2022-05-12
[Feat] [Translations] [Spanish]: config

KnorpelSenf:
Translates:
- the project title and description
- the nav bar
- the search bar


KnorpelSenf:
@roj1512 the CI config makes the deployment preview job only run on Markdown changes. Do you think we should include the config file path into the patterns so that this PQ also gets a preview?

habemuscode:
Translations updated

roj1512 (approved):

PR #359, opened 2022-05-14
Make local links consistent and locale-independent

roj1512:
- Replaced `/` links in `/xx` pages with `../`.
- Replaced `.html` in the links with `.md`, like the majority.


KnorpelSenf (approved):
I haven't checked every single link, but I trust your approach.

This drops absolute links. I think that's the right thing to do. We just have to keep in mind that we should now not deviate from the structure of having every file stored exactly one directory deep. That's probably a good structure anyway, we just have to be aware of it.

PR #362, opened 2022-05-14
Add a note about bot.start call

Loskir:


site/docs/hosting/deno-deploy.md
```diff
@@ -11,7 +11,8 @@ It's just like the other many serverless platforms, but dedicated for Deno apps.
 ## Preparing Your Code
 
 1. Make sure that you have a file which exports your `Bot` object, so that you can import it later to run it.
-2. Create a file named `mod.ts` or `mod.js`, or actually any name you like (but you should be remembering and using this as the main file to deploy), with the following content:
+Note that you shouldn't call `bot.start()` or `bot.init()` in that file.
+3. Create a file named `mod.ts` or `mod.js`, or actually any name you like (but you should be remembering and using this as the main file to deploy), with the following content:
```
Loskir, line 15:
```suggestion
2. Create a file named `mod.ts` or `mod.js`, or actually any name you like (but you should be remembering and using this as the main file to deploy), with the following content:
```

KnorpelSenf:
If people confuse webhooks and polling, we need to make this much clearer in a very different way.

Loskir:
I'm not sure if they do, but that's more of a reminder. One can write that call by accident (myself included :D)

KnorpelSenf:
In that case, we need to:
- mention that they should use webhooks not polling
- say what they should use instead of `start`, namely `webhookCallback`
- link people to the comparison

A reminder works for the people who know it—but people who don't know it yet, will only be confused it they are asked not to use `start`.

KnorpelSenf:
@Loskir what do you say about this?

roj1512 (approved):
Thanks for the contribution!

KnorpelSenf:
@habemuscode is taking some time off. Should we merge and maintain a list of changes that can be translated as soon as he's back, or should we simply wait with this?

roj1512 (approved):

roj1512:
I think we should set a deadline time for translating directly to the PR. After that, we will merge and the translators themselves must see where they were left off. We can make the Notifier bot do this, too. (Telling what PR was merged before getting translated into what languages.)

What do you say regarding this, @KnorpelSenf?

KnorpelSenf:
It would be nice to be able to see which translations are still pending. Otherwise, I think the idea with the time frame sounds like a good compromise between keeping things in sync, but also not slowing down the process.

habemuscode (approved):

PR #363, opened 2022-05-14
Fix language for JavaScript code blocks

Loskir:


KnorpelSenf (approved):

KnorpelSenf (changes requested):
Actually, this needs to be synced to both Spanish and Chinese

KnorpelSenf (approved):

PR #366, opened 2022-05-14
Replace .md with .html in the links on the index pages

roj1512:
Related: https://github.com/vuepress/vuepress-next/issues/893#issuecomment-1126793015


site/docs/es/guide/deployment-types.md
```diff
@@ -95,7 +95,7 @@ ______________                                   _____________
 ```
 
 > Tenga en cuenta que, en realidad, ninguna conexión se mantendría abierta durante horas.
-> Las solicitudes de long polling ienen un tiempo de espera por defecto de 30 segundos (para evitar una serie de [problemas técnicos](https://tools.ietf.org/id/draft-loreto-http-bidirectional-07.md#timeouts)).
+> Las solicitudes de long polling ienen un tiempo de espera por defecto de 30 segundos (para evitar una serie de [problemas técnicos](https://tools.ietf.org/id/draft-loreto-http-bidirectional-07.html#timeouts)).
```
roj1512, line 98: Nice one.

KnorpelSenf:
I do not think that we need to do this. They how I understand https://github.com/vuepress/vuepress-next/issues/893#issuecomment-1126793015, we can use `.md` so that our editor tooling plays nicely with the markdown files, and VuePress will automatically compile the extensions over to `.html`.

roj1512:
So I have got this the wrong way.

KnorpelSenf:
Then again, VuePress themselves do this differently

KnorpelSenf:
So maybe you're right

KnorpelSenf:
No, they only do it on the main page, but no in articles. wtf

roj1512:
Yes only in the `---`

KnorpelSenf:
So should we
1. simply do it the exact same way as them,
2. ask them why it is not consistent (maybe they made a mistake?)
3. leave it in `.md`, or
4. leave change it to `.html` by merging?

roj1512:
I think it is OK to have .md links when we know VuePress redirects the user to it, not by something else. Because if the user clicks a link on the website which is on a native VuePress element, the redirection works.

But we should replace .md links with .html in places like the index page's buttons.

KnorpelSenf:
So that means we go with option 1. I think that's a good idea. If I understand this correctly, then they do indeed transpile `.md` links everywhere except in the frontmatter.

site/docs/advanced/README.md
```diff
@@ -1,5 +1,5 @@
 ---
-next: ./middleware.md
+next: ./middleware.html
```
KnorpelSenf, line 2: Do they do it that way? It looks to me like the extensions are transpiled properly here, too.

roj1512: image

roj1512: I saw this.

KnorpelSenf: I saw 
image
on https://grammy.dev/guide/getting-started.html#getting-started-on-deno which [uses](https://github.com/grammyjs/website/blob/5ea6449081558b4b70933eeced684cac7b9f91bf/site/docs/guide/getting-started.md?plain=1#L2) `.md`.

roj1512: Alrighty.

KnorpelSenf: I think it does not really matter. They use HTML, but MD will be transpiled anyway, so the outcome really is identical. Pick whatever you prefer

roj1512: I was wrong about the documentation. I picked what they prefer.

KnorpelSenf (approved):

PR #370, opened 2022-05-19
Add link to grammY on Twitter

KnorpelSenf:
https://twitter.com/grammy_js


roj1512:
I don't think so.
image

KnorpelSenf:
It says it's also know as 推特, but I am not sure which version is more common. @WingLim?

roj1512:
Even if it is called something else, we haven't done it wrong now, we can replace it if needed anytime.

PR #377, opened 2022-05-22
Fix punctuation on Spanish landing page

KnorpelSenf:


roj1512 (approved):
Thank you. I don't think this requires an approval by @habemuscode. It is as clear as a bell.

PR #382, opened 2022-05-23
Fix active links

KnorpelSenf:
With the current deployment, the links in the navigation menu don't light up if their route is visited. Since we are using the Markdown file extensions which get transpiled to HTML extensions, the routes don't match.

This PQ changes all nav links to use HTML extensions. In addition, it introduces more precise, regex-based matching for links from README files, i.e. overview pages.

### DIFF
--- a/site/docs/.vuepress/config.ts
+++ b/site/docs/.vuepress/config.ts
@@ -50,59 +50,60 @@ export default defineUserConfig({
                 children: [
                   {
                     text: "Overview",
-                    link: "/guide/README.md",
+                    link: "/guide/",
+                    activeMatch: "^/guide/$",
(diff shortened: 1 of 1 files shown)


CikiMomogi:
Is it normal for these two links to be active at the same time?

Screenshot_20220731_204131

KnorpelSenf:
Yes. Both navigation menu entries link to the exact same page, but at different scrolling positions.

However, if this looks bad, we may want to disable that the latter of the two links is ever highlighted.

CikiMomogi:
I was hoping I could fix it by modifying the regex in the `activeMatch` key but it did not work at all.

PR #383, opened 2022-05-23
Update styles for inline code in tables

Loskir:
See /guide/filter-queries.html#query-structure

### DIFF
--- a/site/docs/.vuepress/styles/index.scss
+++ b/site/docs/.vuepress/styles/index.scss
@@ -13,6 +13,7 @@ html.dark {
   --c-brand: #009dca;
 }
 
+// adjust table background colors to avoid clashes with inline code snippets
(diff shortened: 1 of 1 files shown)


site/docs/.vuepress/styles/index.scss
```diff
@@ -13,6 +13,14 @@ html.dark {
   --c-brand: #009dca;
 }
 
+html.dark tr:nth-child(2n) {
```
KnorpelSenf, line 16:
```suggestion
// adjust table background colors to avoid clashes with inline code snippets
html.dark tr:nth-child(2n) {
```

KnorpelSenf (changes requested):

KnorpelSenf (approved):

roj1512 (approved):
Thanks for spotting it!

PR #385, opened 2022-05-24
Deploy the notifier only for pushes to `notifier/`

roj1512:


KnorpelSenf:
How did you make sure that the old way of deploying things stopped? Did you remove the link to the repo, or does it detect that automatically?

roj1512:
I have changed the settings from Deno Deploy.
image

KnorpelSenf (approved):

PR #390, opened 2022-05-24
Remove lost English sentence in Spanish docs

KnorpelSenf:
#379 accidentally introduced hyphens into inline code snippets.
screenshot
This is undesired. This PQ fixes the problem.

In addition, we now permit breaking inline code snippets between words in camelCase.
image

Last but not least, another mistake in the Spanish translations was fixed, where a lost English sentence wasn't removed.


roj1512 (approved):

Loskir (approved):

KnorpelSenf:
That was fast. Pleasure working with you.

PR #392, opened 2022-05-24
Create logos/README.md

roj1512:


KnorpelSenf (commented):
Should the rejected graphic be deleted? I don't think we have a use case for it right now. It will still exist in the git history.

KnorpelSenf (approved):
Thanks!

PR #408, opened 2022-05-27
Use local image files

KnorpelSenf:
No longer fetch the images from Telegram. This will hopefully improve page load speed for this plugin.


KnorpelSenf:
The performance improvements still have potential

KnorpelSenf:
I am still not sure if this will actually help much. The images were reduced in size from 803 KB to 261 KB, but Lighthouse reports the same performance score.

I suggest that we merge this anyway, as the best practices score increased from 92 % to 100 % now. I doubt that Google uses this as a factor. We'll see if it helps.

roj1512 (approved):
The delta of the performance of the process of loading the images is clearly visible.

PR #416, opened 2022-05-30
Fix weird code order

KnorpelSenf:
It looks odd to use the custom context type before declaring it.

Also fixes a mistake in Spanish headers. I really find a lot of those types of mistakes in Spanish recently :(


roj1512 (approved):
Thank you!

PR #417, opened 2022-05-31
Fix variables in Fluent plugin

Loskir:


KnorpelSenf (approved):

KnorpelSenf:
I don't even know how they ended up being like that … thanks for fixing it!

Issue #369, opened 2022-05-19
Indonesian Translation Tracking


CikiMomogi:
# 🇮🇩 Indonesian
## Deskripsi
Issue ini digunakan untuk memantau proses translasi semua halaman. Issue akan ditutup ketika semua dokumentasi berhasil diselesaikan.

### Menjadi Penerjemah
1. Sebelum memulai, **pahami** dulu [panduan berkontribusi untuk penerjemah](https://github.com/grammyjs/website/blob/main/CONTRIBUTING.md#translations).
2. Terjemahkan dengan bahasa yang kamu kuasai. **JANGAN** hanya copas alias copy-paste mentah-mentah dari google translate. Gunakan pemahaman kamu agar menghasilkan terjemahan yang baik dan enak dibaca.
3. Ambil satu halaman yang belum diterjemahkan. Terus terjemahkan. Kalau sudah, buat pull request (PR) sesuai panduan di point satu di atas. Ingat, **satu PR untuk satu halaman** agar mudah dikerjakan dan dikoreksi.

### Menjadi Penyunting
Silahkan dikoreksi PR translasi yang sudah dibuat. Tambahkan catatan jika diperlukan.

## Catatan
Kata-kata yang berhubungan dengan _methods_, nama _function_, _variables_... **tidak perlu** diterjemahkan. Berikut beberapa contohnya:
```
bot
update
keyboard / inline keyboard
webhook
payload
... (dan lain-lain)
```

# 🇬🇧 English
## Description
This issue is for tracking the translation progress of all pages and will be closed when the documentation is finished.

### Being a translator
1. Read [how to contribute in translations](https://github.com/grammyjs/website/blob/main/CONTRIBUTING.md#translations)
2. Translate with your native-speaker level, **DO NOT** just copy the section to some translator tool and paste back the result
3. Take one untranslated page, translate and create PR from your fork to the main project. One PR for page, to better split in parallel the work and less tired reviews

### Being a reviewer
Find the PR's created for translations and review them. Add all notes that you consider.

## Note
Words related to _methods_, _function_ names, _variables_... **mustn't** be translated. Here are some examples:

```
bot
update
keyboard / inline keyboard
webhook
payload
... (feel free to add more, but make sense)
```
# 📝 Translation progress
## README
[...]

PR #418, opened 2022-06-01
Fix duplicate link descriptions

KnorpelSenf:
The content behind the links is redundant, as reported in https://github.com/grammyjs/website/issues/369#issuecomment-1143237068, so the link descriptions are now changed in order to account for that.


roj1512 (approved):
LGTM. Thank you so much!

roj1512 (approved):

roj1512 (approved):

PR #427, opened 2022-06-23
Upload JPEG versions of the WEBP images

dcdunkan:
Broken Instant View for https://grammy.dev/plugins/keyboard.html, because IV does not support WEBP image format.

@KnorpelSenf suggested to serve the JPEG/PNG versions of the WEBP images with them to use in IV.


roj1512 (commented):
Please delete the .webp files and update the links to those images. Thank you for the PR!

roj1512:
Ah

KnorpelSenf (approved):
I'll take the liberty to merge this, it's kinda unrelated to the actual content of the site. All we need to remember is to always provide JPEG equivalents if we ever wanted to upload more files that do not work on IV.

PR #433, opened 2022-07-11
Highlight which config needs to be adjusted in heroku.md

code-withAshish:
The changes in package.json and tsconfig.json neccesary for deploying on heroku were higlighted and comments were added to catch the eye of the reader on instance.
Though not a big update but i propose this change to ease the reader's eyes and give attention to the changes that they need to do instead of going deep into text and sometimes confusing.


code-withAshish:
@KnorpelSenf Please have a look on this PR , it's a small one and might be useful for the documentation

KnorpelSenf:
@roj1512 is maintaining the website, I'll leave the review to him

roj1512 (commented):
I think it is already clear without those comments. If you have more reasons, tell me and I might think again.

site/docs/hosting/heroku.md
```diff
@@ -44,11 +44,11 @@ Our folder structure should now look like this:
 
 After that, open `tsconfig.json` and change it to use this configuration:
 
-```json
+```json{4}
```
KnorpelSenf, line 47: What is this even doing?

code-withAshish: It's highlighting those lines.
Well i just thought it was worth a shot , but yes it's a very small change , may look inappropriate also.
No worries.

CikiMomogi: That's cool! I didn't know the highlight feature existed.

Screenshot Hosting Heroku grammY

KnorpelSenf: Me neither, I like this a lot! I didn't know VuePress was that cool 😄

KnorpelSenf (approved):
I wasn't sure about the PQ at first because you didn't mention the intention behind the change. Now that I understand that it's highlighting the parts of the config that need to be touched, I'm in favour of translating and merging this. It makes it easier for people to see what they need to adjust, rather than comparing all config options on the website and in their file. (Note that some may have made unrelated adjustments which prevents them from just copying over the complete config.)

site/docs/hosting/heroku.md
code-withAshish: lol i thought it's a obvious thing so i didn't do much to explain the problem 
my bad!!!

the whole point was to catch the eye of the user instantly that which part they need to change instead of matching the whole config and maybe making mistakes in between 

but i am glad everyone likes it!!!

roj1512 (approved):

PR #435, opened 2022-07-18
Corrected e.g., comma and, and comma

HeeroML:
Some of these were wrong, e.g., Is typically used with a comma at the end. Also some , and parts are incorrect (two)

### DIFF
--- a/site/docs/guide/basics.md
+++ b/site/docs/guide/basics.md
@@ -6,7 +6,7 @@ next: ./context.md
 # Sending and Receiving Messages
 
 As soon as you start your bot with `bot.start()`, grammY will supply your listeners with the messages that users send to your bot.
-grammY also provides methods to quickly reply to these messages.
+grammY also provides methods to easily reply to these messages.
 
 ## Receiving Messages
 
@@ -57,7 +57,7 @@ Also, check out the [next section](./context.md) to learn how the context object
 
 ## Sending Messages With Reply
 
-You can use the Telegram reply-to feature by specifying the message identifier to reply using `reply_to_message_id`.
+You can use the Telegram reply-to feature by specifying the message identifier to reply to using `reply_to_message_id`.
 
 ```ts
 bot.hears("ping", async (ctx) => {
@@ -117,7 +117,7 @@ File handling is explained in greater depth in [a later section](./files.md#send
 > This can be useful if your bot is running in [privacy mode](https://core.telegram.org/bots#privacy-mode) in group chats.
 
 When you send a message, you can make the user's Telegram client automatically specify the message as a reply.
-That means the user will reply to your bot's message automatically (unless they remove the reply manually).
+That means that the user will reply to your bot's message automatically (unless they remove the reply manually).
 As a result, your bot will receive the user's message even when running in [privacy mode](https://core.telegram.org/bots#privacy-mode) in group chats.
 
 You can force a reply like this:
--- a/site/docs/guide/basics.md
+++ b/site/docs/guide/basics.md
@@ -6,7 +6,7 @@ next: ./context.md
 # Sending and Receiving Messages
 
 As soon as you start your bot with `bot.start()`, grammY will supply your listeners with the messages that users send to your bot.
-grammY also provides methods to easily reply to these messages.
+grammY also provides methods to quickly reply to these messages.
 
 ## Receiving Messages
 
@@ -18,7 +18,7 @@ bot.on("message", (ctx) => {
 });
 ```
 
-However, there are a number of other options, too.
+However, there are several other options, too.
 
 ```ts
 // Handles commands, such as /start.
@@ -28,7 +28,7 @@ bot.command("start", (ctx) => {/* ... */});
 bot.hears(/echo *(.+)?/, (ctx) => {/* ... */});
 ```
 
-You can use auto-complete in your code editor to see all available options, or check out [all methods](https://doc.deno.land/https://deno.land/x/grammy/mod.ts/~/Composer) of the `Composer` class.
+You can use auto-complete in your code editor to see all available options or check out [all methods](https://doc.deno.land/https://deno.land/x/grammy/mod.ts/~/Composer) of the `Composer` class.
 
 > [Read more](./filter-queries.md) about filtering for specific message types with `bot.on()`.
 
@@ -49,15 +49,15 @@ const me = await bot.api.getMe();
 ```
 
 Every method takes an optional options object of type `Other`, which allows you to set further options for your API calls.
-These options objects correspond exactly with the options that you can find in list of methods linked above.
-You can also use auto-complete in your code editor to see all available options, or check out [all methods](https://doc.deno.land/https://deno.land/x/grammy/mod.ts/~/Api) of the `Api` class.
-The rest of this page shows some examples for this.
+These options objects correspond exactly with the options you can find in the list of methods linked above.
+You can also use auto-complete in your code editor to see all available options or check out [all methods](https://doc.deno.land/https://deno.land/x/grammy/mod.ts/~/Api) of the `Api` class.
+The rest of this page shows some examples of this.
 
 Also, check out the [next section](./context.md) to learn how the context object of a listener makes sending messages a breeze!
 
 ## Sending Messages With Reply
 
-You can use the Telegram reply-to feature by specifying the message identifier to reply to using `reply_to_message_id`.
+You can use the Telegram reply-to feature by specifying the message identifier to reply using `reply_to_message_id`.
 
 ```ts
 bot.hears("ping", async (ctx) => {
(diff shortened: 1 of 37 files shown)


KnorpelSenf:
I doubt that putting a comma after _e.g._ is considered correct.

I know for a fact that it's allowed to put a comma before _and_ in order to emphasize that there are different cases, rather than an enumeration.

KnorpelSenf:
In fact, the rule is to never put a comma after e.g. in BrE, but apparently AmE does require one. Since we moved to AmE in #29, this needs to be fixed everywhere (not just in the two places).

HeeroML:
I did it in two places, to start conversation

HeeroML:
I did it in two places to start a conversation. I want to go over every page one by one. So if you are OK with it, I will do some pages today.

HeeroML:
Your right about the comma before and. It's not an error; it's only not needed.

HeeroML:
Sorry, had prettifier on auto when saving :D that's why format check failed so much.

site/docs/guide/deployment-types.md
```diff
@@ -30,15 +30,15 @@ _Imagine you're getting yourself a scoop of ice cream in your trusted ice cream
 You walk up to the employee and ask for your favorite type of ice cream.
 Unfortunately, he lets you know that it is out of stock._
 
-_The next day, you're craving that delicious ice cream again, so you go back to the same place and ask for the same ice cream.
+_The next day, you're craving that delicious ice cream again, so you return to the same place and ask for the same ice cream.
 Good news!
-They restocked over night so you can enjoy your salted caramel ice cream today!
+They are restocked overnight, so you can enjoy your salted caramel ice cream today!
```
KnorpelSenf, line 35: This does not make sense. The employees were not restocked, it was them who did the restocking.

site/docs/guide/deployment-types.md
```diff
@@ -30,15 +30,15 @@ _Imagine you're getting yourself a scoop of ice cream in your trusted ice cream
...
 Yummy._
 
 **Polling** means that grammY proactively sends a request to Telegram, asking for new updates (think: messages).
-If no messages are there, Telegram will return an empty list, indicating that no new messages were sent to your bot since the last time you asked.
+Telegram will return an empty list if no messages are there, indicating that no new messages have been sent to your bot since the last time you asked.
```
KnorpelSenf, line 39: Please revert the former change. It's useful to stick with the convention of listing the condition first when examining all cases.

site/docs/guide/deployment-types.md
```diff
@@ -56,9 +56,9 @@ ______________                                   _____________
 ```
 
 It is immediately obvious that this has some drawbacks.
-Your bot only receives new messages every time it asks, i.e. every few seconds or so.
-To make your bot respond faster, you could just send more requests and not wait as long between them.
-We could for example ask for new messages every millisecond! What could go wrong…
+Your bot only receives new messages every time it asks, i.e., every few seconds or so.
+To make your bot respond faster, you could just send more requests and not wait too long between them.
```
KnorpelSenf, line 60: This is false. We are comparing. It's not about some absolute time interval that should _not be too long_. Please revert.

site/docs/guide/deployment-types.md
```diff
@@ -96,9 +96,9 @@ ______________                                   _____________
 
 > Note that in reality, no connection would be kept open for hours.
 > Long polling requests have a default timeout of 30 seconds (in order to avoid a number of [technical problems](https://tools.ietf.org/id/draft-loreto-http-bidirectional-07.html#timeouts)).
-> If no new messages are returned after this period of time, then the request will be cancelled and resent—but the general concept stays the same.
+> If no new messages are returned after this period, the request will be canceled and resent—but the general concept stays the same.
 
-Using long polling, you don't need to spam Telegram's servers, and still you get new messages immediately!
+You don't need to spam Telegram's servers using long polling, and you still get new messages immediately!
```
KnorpelSenf, line 101: Please revert the former change. It's useful to stick with the convention of listing the condition first when examining all cases.

site/docs/guide/deployment-types.md
```diff
@@ -136,14 +136,14 @@ ______________                                   _____________
...
-Use `bot.start()` and everything will work, no further configuration required.
+Use `bot.start()`, and everything will work; no further configuration is required.
 Under load, you are in complete control of how many messages you can process.
 
 Places where long polling works well include:
 
-- During development on your local machine.
-- On majority of servers.
-- On hosted "backend" instances, i.e. machines that actively run your bot 24/7.
+- During the development of your local machine.
```
KnorpelSenf, line 144: How does this make sense? Why do you suggest that people are developing their local machine? grammY is a framework for bots, not computers. This seems increasingly like you just threw some dumb tool onto the docs without giving any thought to the wording. Maybe I'm wasting my time here?

site/docs/guide/deployment-types.md
```diff
@@ -249,7 +249,7 @@ However, there are a number of drawbacks to using this:
    For instance, they indicate that you always receive a response object, so it is your own responsibility to make sure you're not screwing up while using this minor performance optimization.
 
 If you want to use webhook replies, you can specify the `canUseWebhookReply` option in the `client` option of your `BotConfig` ([API reference](https://doc.deno.land/https://deno.land/x/grammy/mod.ts/~/BotConfig)).
-Pass a function that determines whether or not to use webhook reply for the given request, identified by method.
+Pass a function that determines whether or not to use webhook reply for the given request, identified by the method.
```
KnorpelSenf, line 252: Please revert. It's not about being identified by the method. This is a technical term, and it means _identified by key_, where the key is the method in this case.

site/docs/guide/deployment-types.md
```diff
@@ -332,7 +332,7 @@ You can do this by passing `"return"` as a third argument to `webhookCallback`,
 However, while this behavior has some valid use cases, such a solution usually causes more problems than it solves.
 
 Remember that once you respond to a webhook request, Telegram will send the next update for that chat.
-However, as the old update is still being processed, two updates which were previously processed sequentially, are suddenly processed in parallel.
+However, as the old update is still being processed, two updates that were previously processed sequentially, are suddenly processed in parallel.
```
KnorpelSenf, line 335: Why is this better?

KnorpelSenf (changes requested):

site/docs/guide/basics.md
```diff
@@ -6,7 +6,7 @@ next: ./context.md
 # Sending and Receiving Messages
 
 As soon as you start your bot with `bot.start()`, grammY will supply your listeners with the messages that users send to your bot.
-grammY also provides methods to easily reply to these messages.
+grammY also provides methods to quickly reply to these messages.
```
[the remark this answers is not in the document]
HeeroML:
```suggestion
grammY also provides methods to reply to these messages.
```

site/docs/guide/basics.md
```diff
@@ -18,7 +18,7 @@ bot.on("message", (ctx) => {
 });
 ```
 
-However, there are a number of other options, too.
+However, there are several other options, too.
```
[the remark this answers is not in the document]
HeeroML: Overused Phrase :)

site/docs/guide/basics.md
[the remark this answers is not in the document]
KnorpelSenf: Is there a problem with the word _easy_? The point here is that it makes it simpler to reply to messages. Not faster, and not merely different. Easier.

site/docs/guide/basics.md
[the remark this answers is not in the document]
KnorpelSenf: This is not getting any better by changing all occurrences to _several_, then we just have a different overused phrase.

site/docs/guide/commands.md
```diff
@@ -23,10 +23,10 @@ bot.command("help" /* , ... */);
 bot.command(["a", "b", "c", "d"] /* , ... */);
 ```
 
-Note that only those commands that are in the beginning of a message are handled, so if a user sends `"Please do not send /start to that bot!"`, then your listener will not be called, even though the `/start` command _is_ contained in the message.
+Note that only those commands that are at the beginning of a message are handled, so if a user sends `"Please do not send /start to that bot!"`, then your listener will not be called, even though the `/start` command _is_ contained in the message.
 
-Telegram supports sending targeted commands to bots, i.e. commands that end with `@your_bot_name`.
-grammY handles this automatically for you, so `bot.command("start")` will match messages with `/start` and with `/start@your_bot_name` as commands.
+Telegram supports sending targeted commands to bots, i.e., commands that end with `@your_bot_name`.
+grammY handles this automatically for you, so `bot.command("start")` will match messages with `/start` and `/start@your_bot_name` as commands.
```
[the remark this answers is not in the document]
HeeroML: Then read again. It's a list. One with should be enough. I mean the same and read the same as you say ;)

site/docs/guide/deployment-types.md
HeeroML: I will go through all and revert them all if needed.

site/docs/guide/commands.md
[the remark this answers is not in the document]
KnorpelSenf: Then why are you suggesting to make it less clear?

site/docs/guide/deployment-types.md
KnorpelSenf: This exact case only occurs once in your suggestions.

KnorpelSenf:
Is there still anything going on here? Maybe we should consider closing this due to inactivity.

HeeroML:
You could. I don't know if you guys want to adapt the changes that much later.

KnorpelSenf:
Alright, given that there are a bunch of merge conflicts anyway, I'll close.

PR #436, opened 2022-07-19
specific changes for Mac users

HeeroML:


.gitignore
```diff
@@ -1 +1,2 @@
 /.idea/
+.DS_Store
```
roj1512, line 2: I'm doubtful, is that .DS_Store or .DS_STORE? And why wouldn't it get ignored by default anyway?

HeeroML: i don't know, it didnt and its annoying to remove it every time :D Its, `.DS_Store`

roj1512 (approved):
Well, fine. Let's show some mercy for those appleboys out there.

KnorpelSenf:
What a peculiar change. Is the apple stuff incapable of deactivating these files that pollute git repositores?

HeeroML:
VS Code and git itself add them all the time.

KnorpelSenf:
Aren't those system files?

HeeroML:
Yes

PR #448, opened 2022-07-29
Exclude `a` and `the` from links

KnorpelSenf:
As discussed before https://github.com/grammyjs/website/pull/331#discussion_r933179560

PR #449, opened 2022-07-29
Exclude articles from Spanish links

KnorpelSenf:
Following #448


roj1512 (approved):

roj1512:
I am Indo-European enough to confirm these changes are neat.

KnorpelSenf:
I believe you—I am Indo-European enough to be able to guess my way through autocomplete. I am fine with merging.

PR #455, opened 2022-07-31
Document multi sessions

KnorpelSenf:
Adds docs for https://github.com/grammyjs/grammY/pull/216

### DIFF
--- a/site/docs/plugins/session.md
+++ b/site/docs/plugins/session.md
@@ -335,88 +335,6 @@ In production, you would want to persist your data, for example in a file, a dat
 You should use the `storage` option of the session middleware to connect it to your datastore.
 There may already be a storage adapter written for grammY that you can use (see [below](#known-storage-adapters)), but if not, it usually only takes 5 lines of code to implement one yourself.
 
-## Lazy Sessions
-
-> This section describes a performance optimization that most people do not have to worry about.
-> You may want to continue with the section about [known storage adapters](#known-storage-adapters).
-
-Lazy sessions is an alternative implementation of sessions that can significantly reduce the database traffic of your bot by skipping superfluous read and write operations.
-
-Let's assume that your bot is in a group chat where it does not respond to regular text messages, but only to commands.
-Without sessions, this would happen:
-
-1. Update with new text message is sent to your bot
-2. No handler is invoked, so no action is taken
-3. The middleware completes immediately
-
-As soon as you install (default, strict) sessions, which directly provide the session data on the context object, this happens:
-
-1. Update with new text message is sent to your bot
-2. Session data is loaded from session storage (e.g. database)
-3. No handler is invoked, so no action is taken
-4. Identical session data is written back to session storage
-5. The middleware completes, and has performed a read and a write to the data storage
-
-Depending on the nature of your bot, this may lead to a lot of superfluous reads and writes.
-Lazy sessions allow you to skip steps 2. and 4. if it turns out that no invoked handler needs session data.
-In that case, no data will be read from the data storage, nor written back to it.
-
-This is achieved by intercepting access to `ctx.session`.
-If no handler is invoked, then `ctx.session` will never be accessed.
-Lazy sessions use this as an indicator to prevent database communication.
-
-In practice, instead of having the session data available under `ctx.session`, you will now have _a promise of the session data_ available under `ctx.session`.
-
-```ts
-// Default sessions (strict sessions)
-bot.command("settings", (ctx) => {
-  // `session` is the session data
-  const session = ctx.session;
-});
-
-// Lazy sessions
-bot.command("settings", async (ctx) => {
-  // `promise` is a Promise of the session data, and
-  const promise = ctx.session;
-  // `session` is the session data
-  const session = await ctx.session;
-});
-```
-
-If you never access `ctx.session`, no operations will be performed, but as soon as you access the `session` property on the context object, the read operation will be triggered.
-If you never trigger the read (or directly assign a new value to `ctx.session`), we know that we also won't need to write any data back, because there is no way it could have been altered.
-Consequently, we skip the write operation, too.
-As a result, we achieve minimal read and write operations, but you can use session almost identical to before, just with a few `async` and `await` keywords mixed into your code.
-
-So what is necessary to use lazy sessions instead of the default (strict) ones?
-You mainly have to do three things:
-
-1. Flavor your context with `LazySessionFlavor` instead of `SessionFlavor`.
-   They work the same way, just that `ctx.session` is wrapped inside a promise for the lazy variant.
-2. Use `lazySession` instead of `session` to register your session middleware.
-3. Always put an inline `await ctx.session` instead of `ctx.session` everywhere in your middleware, for both reads and writes.
-   Don't worry: you can `await` the promise with your session data as many times as you want, but you will always refer to the same value, so there are never going to be duplicate reads for an update.
-
-Note that with lazy sessions, you can assign both objects and promises of objects to `ctx.session`.
-If you set `ctx.session` to be a promise, it will be `await`ed before writing the data back to the data storage.
-This would allow for the following code:
-
-```ts
-bot.command("reset", (ctx) => {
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
@Loskir you were asking about these docs, have a look at the preview and navigate to `/plugins/session`

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+
+```ts
+interface SessionData {
+  foo: { prop: number };
+  bar: "empty" | "half" | "full";
+  baz: { width?: number; height?: number };
+}
+```
+
+You can then keep on using `SessionFlavor<SessionData>` for your context object.
```
roj1512, line 606:
```suggestion
You can then keep using `SessionFlavor<SessionData>` for your context object.
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+
+## Lazy Sessions
+
+> This section describes a performance optimization that most people do not have to worry about.
+
+Lazy sessions is an alternative implementation of sessions that can significantly reduce the database traffic of your bot by skipping superfluous read and write operations.
```
roj1512, line 612:
```suggestion
Lazy Sessions is an alternative implementation of sessions that can significantly reduce the database traffic of your bot by skipping superfluous read and write operations.
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+
+Let's assume that your bot is in a group chat where it does not respond to regular text messages, but only to commands.
+Without sessions, this would happen:
+
+1. Update with new text message is sent to your bot
+2. No handler is invoked, so no action is taken
+3. The middleware completes immediately
```
roj1512, line 619:
```suggestion
1. Update with new text message is sent to your bot.
2. No handler is invoked, so no action is taken.
3. The middleware completes immediately.
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+
+As soon as you install (default, strict) sessions, which directly provide the session data on the context object, this happens:
+
+1. Update with new text message is sent to your bot
+2. Session data is loaded from session storage (e.g. database)
+3. No handler is invoked, so no action is taken
+4. Identical session data is written back to session storage
+5. The middleware completes, and has performed a read and a write to the data storage
```
roj1512, line 627:
```suggestion
1. Update with new text message is sent to your bot.
2. Session data is loaded from session storage (e.g. database).
3. No handler is invoked, so no action is taken.
4. Identical session data is written back to session storage.
5. The middleware completes, and has performed a read and a write to the data storage.
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+
+Depending on the nature of your bot, this may lead to a lot of superfluous reads and writes.
+Lazy sessions allow you to skip steps 2. and 4. if it turns out that no invoked handler needs session data.
+In that case, no data will be read from the data storage, nor written back to it.
+
+This is achieved by intercepting access to `ctx.session`.
+If no handler is invoked, then `ctx.session` will never be accessed.
+Lazy sessions use this as an indicator to prevent database communication.
```
roj1512, line 635:
```suggestion
Lazy Sessions use this as an indicator to prevent database communication.
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+As soon as you install (default, strict) sessions, which directly provide the session data on the context object, this happens:
+
+1. Update with new text message is sent to your bot
+2. Session data is loaded from session storage (e.g. database)
+3. No handler is invoked, so no action is taken
+4. Identical session data is written back to session storage
+5. The middleware completes, and has performed a read and a write to the data storage
+
+Depending on the nature of your bot, this may lead to a lot of superfluous reads and writes.
+Lazy sessions allow you to skip steps 2. and 4. if it turns out that no invoked handler needs session data.
```
roj1512, line 630:
```suggestion
Lazy Sessions allow you to skip steps 2. and 4. if it turns out that no invoked handler needs session data.
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+  const session = await ctx.session;
+});
+```
+
+If you never access `ctx.session`, no operations will be performed, but as soon as you access the `session` property on the context object, the read operation will be triggered.
+If you never trigger the read (or directly assign a new value to `ctx.session`), we know that we also won't need to write any data back, because there is no way it could have been altered.
+Consequently, we skip the write operation, too.
+As a result, we achieve minimal read and write operations, but you can use session almost identical to before, just with a few `async` and `await` keywords mixed into your code.
+
+So what is necessary to use lazy sessions instead of the default (strict) ones?
```
roj1512, line 660:
```suggestion
So what is necessary to use Lazy Sessions instead of the default (strict) ones?
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+Lazy sessions is an alternative implementation of sessions that can significantly reduce the database traffic of your bot by skipping superfluous read and write operations.
+
+Let's assume that your bot is in a group chat where it does not respond to regular text messages, but only to commands.
+Without sessions, this would happen:
+
+1. Update with new text message is sent to your bot
+2. No handler is invoked, so no action is taken
+3. The middleware completes immediately
+
+As soon as you install (default, strict) sessions, which directly provide the session data on the context object, this happens:
```
roj1512, line 621:
```suggestion
As soon as you install default (strict) sessions, which directly provide the session data on the context object, this happens:
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+As a result, we achieve minimal read and write operations, but you can use session almost identical to before, just with a few `async` and `await` keywords mixed into your code.
+
+So what is necessary to use lazy sessions instead of the default (strict) ones?
+You mainly have to do three things:
+
+1. Flavor your context with `LazySessionFlavor` instead of `SessionFlavor`.
+   They work the same way, just that `ctx.session` is wrapped inside a promise for the lazy variant.
+2. Use `lazySession` instead of `session` to register your session middleware.
+3. Always put an inline `await ctx.session` instead of `ctx.session` everywhere in your middleware, for both reads and writes.
+   Don't worry: you can `await` the promise with your session data as many times as you want, but you will always refer to the same value, so there are never going to be duplicate reads for an update.
```
roj1512, line 667:
```suggestion
   Don't worry: You can `await` the promise with your session data as many times as you want, but you will always refer to the same value, so they are never going to be duplicate reads for an update.
```

site/docs/plugins/session.md
```diff
@@ -636,6 +554,286 @@ For example, the storage adapter for Supabase can be imported from `https://deno
...
+
+```ts
+interface SessionData {
+  petNames: string[];
+  petBirthdays?: number[];
+}
+```
+
+This would not break your existing session data.
+However, this is not so great, because now the names and the birthdays are stored in different places.
```
roj1512, line 758:
```suggestion
However, this is not so great, because the names and the birthdays are now stored in different places.
```

site/docs/plugins/session.md
KnorpelSenf: There are not going to be duplicate reads. Duplicate reads will not happen.

KnorpelSenf: Shoud we revert?

roj1512 (approved):

roj1512 (approved):

PR #463, opened 2022-08-02
Fix examples repository links

roj1512:


KnorpelSenf (approved):
Thaaaanks

KnorpelSenf:
GitHub refined isn't there yet, it seems

roj1512:
Yes, it is buggy

roj1512:
I don't know why it is like that.

PR #474, opened 2022-08-05
Fix typo and make title case consistent

dcdunkan:


roj1512 (approved):
LGTM. Thank you so much!

PR #478, opened 2022-08-14
fix: asynchronous enter into conversation

mzoterov:


KnorpelSenf:
@roj1512 if we do this here, we need to do it everywhere. Are you in favour of that?

roj1512:
No.

KnorpelSenf:
Given that the existing code example is correct, I will close this. Feel free to comment or repoen.

PR #492, opened 2022-09-04
Include deno.land/std in current-versions plugin

dcdunkan:


roj1512 (approved):
LGTM. Thank you so much!

KnorpelSenf:
`std` regularly has breaking changes which is why we initially omitted it from the plugin. Are we sure that updates to `std` cannot break our examples? How will this be tested?

roj1512:
There's no breaking changes in the std, nor the Deno API, unless the module is using an unstable API.

Source: https://twitter.com/biwanczuk/status/1565305775743983624

KnorpelSenf:
Great! Then this makes sense, thanks for the link

PR #499, opened 2022-09-06
Update dependencies to fix vulnerabilities

roj1512:


roj1512:
I'm not sure if we will be safe.

KnorpelSenf:
What does that mean?

roj1512:
It means I'm not sure if doing that breaks things or not.

KnorpelSenf:
No one knows this until we try, that's why I'm suggesting to give it a shot :)
If it does break, we will see it. Then we can revert and report the bug etc

roj1512:
@KnorpelSenf Vulnerabilities can't be fixed if we upgrade the dependencies.

KnorpelSenf:
Why not?

roj1512:
Try it yourself.

roj1512:
@KnorpelSenf This is not a draft, you're trying to change its goal.

KnorpelSenf:
That's a vuln in the deps. Usually updating deps helps, and we need to do this anyway again, so why not do it now? The goal is to be up to date so that we avoid vulns.

Draft because there are pending discussions that need to be resolved, so this isn't ready to merge.

roj1512:
Pending discussions could be in a different PR that does the dependency upgrades. I thought you were just meaning to remove the vulnerability alerts at first.

KnorpelSenf:
I was, but in the most future-proof way, which would be not to rely on months-old beta versions. Do you feel like this needs to be merged fast?

roj1512:
Well, that's a thing, too. The warning can be nothing.

roj1512:
Drafting it again

KnorpelSenf:
> @KnorpelSenf Vulnerabilities can't be fixed if we upgrade the dependencies.

I do not understand what you mean. It seems to me like all vulnerabilities are fixed if VuePress is updated to the most recent version. In addition, some bugs are fixed. For example, the production site does not remove the header anchor from the URL when scrolling to the top.

KnorpelSenf:
```sh
$ npm audit
found 0 vulnerabilities
```

KnorpelSenf (approved):

PR #500, opened 2022-09-06
add preliminary support for shiki-twoslash

roziscoding:


KnorpelSenf:
We are having the formatting problem again. Could you run `deno fmt` here?

KnorpelSenf:
CI fails due to OOM. We cannot build and publish the site like that. Do you have any idea how to fix this?

roziscoding:
> CI fails due to OOM. We cannot build and publish the site like that. Do you have any idea how to fix this?

I was hoping you wold know 😅

But I'll take a look later today

roziscoding:
It's weird. Local build works just fine 🤔

roziscoding:
It appears the vite build is indeed very memory hungry, and adding shiki-twoslash to that mix increased it even more.
I've tryed every solution pointed in https://github.com/vitejs/vite/issues/2433, but the only thing that really worked was to increase the heap size limit
It's ugly, but does get the job done

KnorpelSenf (commented):
This already looks pretty great!

1. I don't mind that we need to use more RAM. It's a constant memory consumption, and not something that explodes due to bad algorithms.
2. That's no Markdown rendering in the code tooltips. I assume this isn't easily possible?
3. I would prefer to update the entire website and all code examples at the same time, rather than migrating slowly and being in an inconsistent state for a long time. It may be better to open a new PQ for that. Therefore I'd merge the config only from here.

KnorpelSenf:
Well actually we can also just use this PQ for the migration

roziscoding:
> 2. That's no Markdown rendering in the code tooltips. I assume this isn't easily possible?

I haven't found an option to enable this with shiki, but it should be possible somehow. We could, for example, get the highlighted shiki code, extract the tooltips (they most likely have a class name) and pass it manually through markdown-it.

> 3. I would prefer to update the entire website and all code examples at the same time, rather than migrating slowly and being in an inconsistent state for a long time. It may be better to open a new PQ for that. Therefore I'd merge the config only from here.

I have one concern about this, which is build time.
While working on getting things to work, I implemented this in a way that enabled twoslash for every code block, unless opted out. When I did that, tho, the vuepress build process just froze. I didn't wait to see how much time it'd take to finish, but it is enough to raise the concern.
Though, like we talked about in the telegram group, not every code block will actually benefit from this, so maybe we enable it only when it makes sense?

KnorpelSenf:
We'll have to try this. Usually, the process should never freeze completely, as there's always logging output.

KnorpelSenf:
> We could, for example, get the highlighted shiki code, extract the tooltips (they most likely have a class name) and pass it manually through markdown-it.

This sounds like too much work. I'd just leave it for now.

KnorpelSenf:
I don't really see this happening anytime soon, and I also don't think that a few type annotation are worth the complexity that it would take to get this through. Closing.

If you ever feel like picking it up again, feel free to reopen.

PR #501, opened 2022-09-12
Suggest global placeables docs

KnorpelSenf:
As requested in https://github.com/grammyjs/i18n/pull/20#issuecomment-1243195196. Notice the target branch.


dcdunkan (approved):
Thank you!

PR #503, opened 2022-09-16
Make the `ratelimiter` name casing consistent

CikiMomogi:
Applies consistent casing of the plugin name. See https://github.com/grammyjs/ratelimiter/pull/4.


roj1512 (approved):
These changes look good to me. Although, generally speaking, the documentation isn't really solid.

PR #523, opened 2022-10-09
feat: create SetWebhookUtil component

### LINKED ISSUE #517: feat: `SetWebhookUtil` component
It would be awesome to have a tiny utility component which can be included in all the hosting guides which require a webhook to be set. While it is not hard for many users to correctly set the webhook URL using `curl` or other tools, some beginners struggle with understanding which URL to fetch.

It would be nice to have a small interactive component on the website which lets users paste their token and URL and then it will set the webhook for them. It could also get some neat information to validate the info:

1. Call `getMe` with the token to confirm the bot name
2. Pass the URL to the `URL` constructor to validate the URL
3. Display information from `getWebhookInfo` so that there is an easy way to confirm that things are working correctly
4. Provide a tooltip that explains that the bot token will never be used for anything else than setting for this component to function, and tell users to never give away their token to anyone, no matter how polite a website asks.
5. Have a link that opens the chat with the new bot in a new tab so that users can easily head over to Telegram when the webhook was set and the bot is working.
6. Perhaps have a second button to delete the webhook again? This might not be necessary, but it could be that it feels more natural to have it. Sometimes people feel the urge to clean up after themselves.
7. Perhaps store the token locally in the browser so that it persists across page loads? We must not do this if it feels shady, I don't want anyone to believe we store the token server-side (even though we will never do this)

roziscoding:
Preview:

image
image
image
image

Closes #517


KnorpelSenf:
Would you be able to defer loading everything until hydration kicks in, using a dynamic import? I'm not sure it's a good idea to bundle the component framework and grammY and everything else into the static site, this will add a few hundred kB to the build.

The component is usually at the very bottom of the page, so in 99 % of page visits, we have a few minutes to download the extra weight and render the component.

Yes, this will mean that the component doesn't load if JS is disabled, but this actually makes a lot of sense anyway.

roziscoding:
All features are now implemented.
The build fails locally because I didn't use the CDN version of grammY. I'll need some help getting that done.
Once I can build a production version, I can test possible performance improvements.

KnorpelSenf:
It looks like it OOMs?

KnorpelSenf:
I can take a closer look at a later point

roziscoding:
> It looks like it OOMs?

Yeah, in github actions it fails due to OOM, but locally the error is different.
I'll post it here in about an hour when I get to the computer.

KnorpelSenf (commented):
Nice to finally see this unblocked 🤓
Ping me when I should review

roziscoding:
Closing this in favor of https://github.com/grammyjs/website/pull/633

PR #504, opened 2022-09-18
Add section about installing conversations with custom session data

KnorpelSenf:
As discussed in https://t.me/grammyjs/78172 and asked many times before.

PR #525, opened 2022-10-13
Adjust time to translate in guide

KnorpelSenf:
2-3 days isn't realistic anymore, see #504 which has 2/3 translations after 18 days.


roj1512 (approved):
LGTM. Thank you so much!

PR #527, opened 2022-10-14
Add more details about multiple listeners in same command

DaviSMoura:


KnorpelSenf:
I'm in favour of these changes. People have run into this problem repeatedly, and while the part about middleware technically documents this already, the earlier sections about message handling are somewhat assuming that the reader skipped ahead and had knowledge of a later part. That's undesired.

This needs to be documented in more places, not just for commands. This definitely should be added to filter queries. What do you think about adding it to the part about receiving messages, too? It seems to me like people might not care about that at this early point yet, right?

KnorpelSenf:
Closing due to inactivity. Feel free to comment or reopen if you're willing to pick up development again :)

PR #528, opened 2022-10-15
Fix text justify

CikiMomogi:
Oops!


KnorpelSenf (approved):
Whoopsie how did this slip through the review process? 🥲
Thanks for the fix!

KnorpelSenf:
@roj1512 I took the liberty and merged without consulting you because it was obviously the right change to perform

roj1512:
That was what I was expecting

PR #533, opened 2022-10-26
Refactor

CikiMomogi:


.gitignore
```diff
@@ -1,2 +1,6 @@
 /.idea/
 .DS_Store
+
+#pnpm
+site/pnpm-lock.yaml
+#pnpm end
```
roj1512, line 6: I think these changes should be kept only on your side since we are neither using pnpm, nor we have suggested it anywhere.

CikiMomogi: I don't think it's a big deal to include them. That can make it easier for use to exclude the file every time we want to commit.

roj1512: Why would you need to exclude it every time you commit? Just don't include it, that's it.

CikiMomogi: Correct. Everytime i want to commit, i must exclude them manually.

roj1512: What? How is that possible?

roj1512: As far as I know, Git does not add any changes, from any file, unless you explicitly do it with `git add`. Do you always do `git add -A`?

CikiMomogi: Yup. I always do git add -A rather than add every files manually. But, it is not big deal thou. Should i remove this line?

roj1512: I don’t mind if it stays there. I was just curious about it. It doesn’t change anything for the end users.

CikiMomogi: Trust me. It will help me a lot. 😁

roj1512: Well, let’s keep it then :)

roj1512 (changes requested):
We should add a preceding `-` or `_` to the names of the imported `.sass` files.

roj1512 (approved):
Thanks! Sorry for the delay.

PR #534, opened 2022-11-05
fix(conversation): promisify active conversation

waptik:
without the fix, you get the following `[AsyncFunction: active]`


KnorpelSenf:
Oops sorry, I did not see this one before fixing it up in #535. The translations are already done there, so I'll close this in favour of #535. Thanks for contributing, though!

PR #537, opened 2022-11-07
Okay okay fine we'll update it

KnorpelSenf:
Just for you @CikiMomogi <3

PR #538, opened 2022-11-07
Okay okay fine we'll update it

roj1512:
Mistakenly merged before syncing with all languages #537

Just for you @CikiMomogi <3


habemuscode (approved):
Can be merged since ID changes is the only language required here

PR #539, opened 2022-11-15
Fix broken IETF link

KnorpelSenf:
No idea why the document was removed?


roj1512 (approved):
LGTM. Thank you so much!

PR #541, opened 2022-11-16
Update notifier deps

roj1512:


notifier/import_map.json
```diff
@@ -0,0 +1,11 @@
...
+  "imports": {
+    "envalid": "https://deno.land/x/envalid@0.1.2/mod.ts",
+    "grammy": "https://deno.land/x/grammy@v1.11.2/mod.ts",
+    "oak": "https://deno.land/x/oak@v11.1.0/mod.ts",
+    "oak/": "https://deno.land/x/oak@v11.1.0/",
+    "octokit/": "https://raw.githubusercontent.com/octokit/webhooks-methods.js/v2.0.0/src/",
+    "postgres": "https://deno.land/x/postgres@v0.17.0/mod.ts",
+    "std/": "https://deno.land/std@0.164.0/"
+  }
+}
```
KnorpelSenf, line 11: Consider being consistent here by never including the file path to entrypoints in the import map.

KnorpelSenf (approved):
LGTM, just a suggestion that is fully optional:

notifier/import_map.json
roj1512: Do you by entrypoint mean "/"?

roj1512: If I don't do "std/", neither "testing/", I will waste minutes in thinking of a good name for "testing/asserts.ts".

KnorpelSenf: I mean that `/mod.ts` could be moved to the import statements, rather than being hard-coded here.

KnorpelSenf (approved):
so pretty, such wow

PR #546, opened 2022-11-26
Fix typo

abdoo9:


KnorpelSenf (approved):
Thanks! Nice catch

roj1512 (approved):
LGTM. Thank you so much!

KnorpelSenf:
@roj1512 does CI fail because the commit on main does not have the correct labels? Does this mean we should add a check for the branch?

KnorpelSenf:
No wait, it succeeds on main … what's going on here

roj1512:
No idea.

roj1512:
I think we should debug it.

PR #547, opened 2022-11-26
Remove Chinese translations

roj1512:
This is due to the inactivity of the contributors that might lead to unwanted partial and/or out-of-date translations.


KnorpelSenf:
As sad as I am to see this language removed, I must admit that the missing Chinese translations are blocking our workflow. There are many PQs that are many weeks old and are otherwise fully translated.

If @WingLim is busy with other things, that is fine. We are in no way interested in forcing anyone to work on a project even though they have largely moved on.

If would be very happy to add Chinese translations again as soon as the translations are catching up again! The git history will preserve the existing work.

KnorpelSenf (changes requested):
Please add a note to the translation guide with a link to the last commit on `main` which contains Chinese translations. This will help any potential future translators to start off more easily.

KnorpelSenf (approved):

PR #551, opened 2022-11-27
Apply commits before Chinese was removed

### LINKED ISSUE #548: Wrong commit order for restoring Chinese
Before #547 was merged, there were a number of PQs that do not include Chinese translations but still change all other languages.

It would be great to have a single commit to put into `TRANSLATING.md` which marks the point at which Chinese was removed. If anyone wants to pick it up, all they have to do is to look what changed since that commit. With the current history, it would be 0ef23b667390abcc6036497e70fe17761bc76020 but then the commit range 48aa7020c44b7975d844f59a18371b965eaf93e6..d4f552c4c1c65a11389ba14e7f096be38a3bc586 (which happened before) would also have to be applied, as the translations are missing here.

A good strategy to mitigate this is to create 3 PQs:
- [x] restore the repo contents at c2aeeb64416de4c216d7317fd28c17a099d06f18 (#549)
- [x] apply 0ef23b667390abcc6036497e70fe17761bc76020 from #547 (#550)
- [x] apply all of 48aa7020c44b7975d844f59a18371b965eaf93e6..d4f552c4c1c65a11389ba14e7f096be38a3bc586 (#551)

KnorpelSenf:
Closes #548.

This applies the commit range 48aa7020c44b7975d844f59a18371b965eaf93e6..d4f552c4c1c65a11389ba14e7f096be38a3bc586


roj1512 (approved):
Thanks!

PR #554, opened 2022-11-28
Improve translations and fix inconsistencies

habemuscode:


KnorpelSenf:
I guess if it's Spanish-only then it shouldn't be translated

roj1512 (approved):

PR #559, opened 2022-12-05
Inform about silent supergroup migrations

Madnex:
I think it would be worth mentioning that telegram groups are migrated automatically behind the scenes to supergroups under certain conditions. If not aware this can lead to some unexpected behavior of the session plugin.


KnorpelSenf (commented):
Good idea (we should mention this problem) but flawed execution. We should
- offer a solution or a workaround
- explain why no solution exists

Yes, that's a contradiction, and yes, that's the reason why we didn't explain this earlier.

The two chats are completely separate, so what are we going to do about this?

(Also, I'd like to improve on structure, style and orthography before merging—but we need to answer the above question first, anyway.)

Madnex:
I believe if we want to share the same session data for both chats it is required to somehow store the link between the two chat ids. However, to my knowledge this link is only temporarily provided by telegram. The [message type](https://core.telegram.org/bots/api#message) has two properties that contain the link info `migrate_to_chat_id` and `migrate_from_chat_id`. 

This in combination with the `chat.id` of the update gives the link between the old group and the new super group. When I experimented with the migration behavior I could only get these properties (`migrate_to_chat_id` and `migrate_from_chat_id`) once though. Don't know how this is designed but for sure you can not access this information for all future messages.

Once this information is catched by the bot I think one solution could be to add a value (e.g. `migrated_to`) for the old session object (old key) that contains the new key. And additionally a new session object (with the new key) with the information for the old session (e.g. `migrated_from`). And then always when loading session data check for the `migrated_from` field and point to that one instead, i.e. the new session object is only a pointer towards the old session.

What do you think? Or would there be a reason to maintain two sessions for both chat ids and then apply some other logic to combine them later on?

btw: should we continue this here or create an issue for this discussion?

KnorpelSenf:
There is a way to link one chat to the other, which is what you outlined. However, this doesn't mean that automatic session migrations from one chat to the other are possible. This is because new messages can be received by the bot from both chats concurrently for a short period of time. This leads to data races which cannot be fixed without help from the bot developer. That's what I meant when I said that we have to

> * explain why no solution exists

KnorpelSenf:
How should we proceed with this?

Madnex:
Do you have more official information on this? I can try to put this conversation into a comprehensive text, explaining the two bullet points. But I can't find much info on this topic so I'm not certain about what exactly to explain.

KnorpelSenf:
Here is the Bot API dev himself confirming the data races: https://t.me/tdlibchat/44915

Madnex:
> Here is the Bot API dev himself confirming the data races: https://t.me/tdlibchat/44915

Thanks for the info. Do I understand correctly that we can hope that this will be fixed eventually (https://t.me/tdlibchat/42933)?

KnorpelSenf:
> > Here is the Bot API dev himself confirming the data races: https://t.me/tdlibchat/44915
> 
> Thanks for the info. Do I understand correctly that we can hope that this will be fixed eventually (https://t.me/tdlibchat/42933)?

No. Read the respective message thread, it continued with https://t.me/tdlibchat/44906.

Madnex:
Okay, I see. Added some more explanation about this problematic with two possible (inconsistent workarounds). Should we put more details or keep it on this level?

KnorpelSenf (changes requested):
I think this level of detail is good, I like your changes.

There are still a few things that could be improved, though:

1. You describe the issue in a warning box, and then start a new section to suggest the workarounds. It makes more sense to explain the problem and the solution in one. Perhaps we could name the section just Supergroup Migrations or something, and then do all the explaining.
2. Do you think it would better in its own section, rather than putting it alongside the session keys? On the one side it is related to session keys, but on the other side it's still somewhat of a rare edge case. It makes sense to provide readers with the important information first (how to connect to a storage?), and when their bot is running, we can tell them about these edge cases to harden the code base (what if there are chat migrations?). I'm a little undecided here, what do you think?
3. You currently mention two workarounds: doing unreliable chat migrations, or disabling sessions for regular groups. There are more ways to go about this:
    - Ignore the problem. The bot's session data will effectively reset when a group is migrated. Simple, reliable, default behaviour, but potentially unexpected once per chat.
    - Only storing temporary data (or data with timeouts) in the session, and using a database for the important things that need to be migrated when a chat migrates. This can then use transactions and custom logic to handle concurrent data access from the old and the new chat. A lot of effort and has a performance cost, but the only truly reliable way to solve this problem.
    - Letting the users decide explicitly. (“This chat was migrated, do you want to transfer the bot data?”) Much more reliable and transparent than automatic migrations due to the artificially added delay, but worse UX.
4. Try keeping it at one sentence per line, as it is explained in the contribution guide in this repo. (There might be a few more nitpicks about style later on because for some reason we all care a bit too strongly about excellent language here, but it won't be much effort on your side, so don't worry about that for now.)


Thanks for your efforts! Looking forward to getting this ready for merge :)

Madnex:
Thanks for the detailed feedback! 

For 1. and 2.: Yes, this is something I was wondering as well. I think it's worth making it a section. For the placement within the documentation: Is there any page / section about general caveats / good to know things? Something a bit like in the "Advanced/[Deployment Checklist](https://grammy.dev/advanced/deployment.html#deployment-checklist)" but more general. If not, maybe it's worth opening a new page? Or just in the [Overview page](https://grammy.dev/advanced/#overview-advanced-topics)? 

Do you have other caveats / tips in mind that are worth noting? Maybe also just general Telegram behavior that one might stumble over while developing a bot. This could all go to the decided place then.

And then in the [Session Page](https://grammy.dev/plugins/session.html#session-keys) we just put a short warning box linking to the explanation.

The suggested other workarounds are added in the new commit. If the content is fine like this, it is now just a matter of finding a place for it and prettifying the language / style :)

site/docs/plugins/session.md
```diff
@@ -326,6 +326,29 @@ If you must use the option (which is of course still possible), you should know
...
+
+This migration can only occur once per group chat but it can cause inconsistencies. This is because the migrated chat is technically a completely different chat with a different `chat.id` and hence a different session id.
+
+Currently there is no safe solution to this problem because the two chats can send messages simultaneously leading to data races. However, there are several ways of dealing with this issue:
+
+* Ignoring the problem. The bot's session data will effectively reset when a group is migrated. Simple, reliable, default behavior, but potentially unexpected once per chat. For example if a migration happens while a user is in a conversation (the conversation plugin uses sessions) the conversation will be reset.
+
+* Only storing temporary data (or data with timeouts) in the session, and using a database for the important things that need to be migrated when a chat migrates. This can then use transactions and custom logic to handle concurrent data access from the old and the new chat. A lot of effort and has a performance cost, but the only truly reliable way to solve this problem.
```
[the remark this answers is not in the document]
KnorpelSenf: @roj1512 what's your opinion on this?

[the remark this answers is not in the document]
Madnex: I added two sentences in [#61b72f1](https://github.com/grammyjs/website/pull/559/commits/61b72f14fdbec4f4a20be58606056a9e123f8974). Maybe this puts it into context a bit?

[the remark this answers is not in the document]
KnorpelSenf: Looking good, thanks

roj1512:
Nothing is remaining except my take over.

KnorpelSenf:
Awesome!

Madnex:
> @Madnex could you fix the formatting issues?

What exactly do I need to change? I don't see where the format breaks?

KnorpelSenf:
Check the CI task

KnorpelSenf:
You can run the formatting script to fix it automatically :)

KnorpelSenf:
See https://github.com/grammyjs/website#checking-correct-formatting-of-docs

Madnex:
Thanks, got it I think 👍

KnorpelSenf:
Awesome, thanks!

KnorpelSenf (approved):
LGTM, ready for translation IMO

site/docs/plugins/session.md
```diff
@@ -326,6 +326,46 @@ If you must use the option (which is of course still possible), you should know
 Make sure you understand the consequences of this configuration by reading [this](../guide/deployment-types.md) article and especially [this](./runner.md#sequential-processing-where-necessary) one.
 :::
 
+### Chat Migrations
+
+If you are using sessions for a group chat you should be aware that Telegram migrates regular groups to supergroups under certain circumstances (e.g. [here](https://github.com/telegramdesktop/tdesktop/issues/5593)).
```
roj1512, line 331:
```suggestion
If you are using sessions for groups, you should be aware that Telegram migrates regular groups to supergroups under certain circumstances (e.g. [here](https://github.com/telegramdesktop/tdesktop/issues/5593)).
```

site/docs/plugins/session.md
```diff
@@ -326,6 +326,46 @@ If you must use the option (which is of course still possible), you should know
...
+
+This migration can only occur once per group chat but it can cause inconsistencies.
```
roj1512, line 333:
```suggestion
This migration only occurs once for each group, but it can cause inconsistencies.
```

site/docs/plugins/session.md
```diff
@@ -326,6 +326,46 @@ If you must use the option (which is of course still possible), you should know
 Make sure you understand the consequences of this configuration by reading [this](../guide/deployment-types.md) article and especially [this](./runner.md#sequential-processing-where-necessary) one.
 :::
 
+### Chat Migrations
+
+If you are using sessions for groups, you should be aware that Telegram migrates regular groups to supergroups under certain circumstances (e.g. [here](https://github.com/telegramdesktop/tdesktop/issues/5593)).
+
+This migration can only occur once per group chat but it can cause inconsistencies.
+This is because the migrated chat is technically a completely different chat with a different chat identifier and hence a different session identifier.
```
roj1512, line 334:
```suggestion
This is because the migrated chat is technically a completely different chat that has a different chat identifier, and hence a different session identifier.
```

site/docs/plugins/session.md
```diff
@@ -326,6 +326,46 @@ If you must use the option (which is of course still possible), you should know
...
+
+Currently, there is no safe solution to this problem because the two chats can send messages simultaneously.
```
roj1512, line 336:
```suggestion
Currently, there is no safe solution for this problem because messages from the two chats are also differently identified.
```

site/docs/plugins/session.md
```diff
@@ -326,6 +326,46 @@ If you must use the option (which is of course still possible), you should know
...
+This can lead to data races.
+However, there are several ways of dealing with this issue:
+
+- Ignoring the problem.
+  The bot's session data will effectively reset when a group is migrated.
+  Simple, reliable, default behavior, but potentially unexpected once per chat.
+  For example if a migration happens while a user is in a conversation (the conversation plugin uses sessions) the conversation will be reset.
```
roj1512, line 343:
```suggestion
  For example, if a migration happens while a user is in a conversation powered by the [conversations plugin](./conversations.md), the conversation will be reset.
```

site/docs/plugins/session.md
```diff
@@ -326,6 +326,46 @@ If you must use the option (which is of course still possible), you should know
 Make sure you understand the consequences of this configuration by reading [this](../guide/deployment-types.md) article and especially [this](./runner.md#sequential-processing-where-necessary) one.
 :::
 
+### Chat Migrations
+
+If you are using sessions for groups, you should be aware that Telegram migrates regular groups to supergroups under certain circumstances (e.g. [here](https://github.com/telegramdesktop/tdesktop/issues/5593)).
```
roj1512, line 331: How is this Telegram Desktop issue related?

Madnex: It's just the only example / documentation for the impact of the migration I could find online. It's just an example of how to distinguish supergroups from normal groups (the part about "Actual behaviour"). Nothing in particular about Telegram Desktop.

roj1512: Still a little misleading, don’t you think?

Madnex: Well, I don't know. Tbh I did not even notice that this is about Telegram Desktop. We could also use [this link](https://stackoverflow.com/questions/54555821/why-telegram-group-automatically-becomes-super-group) instead

roj1512: Can you remove the link?

KnorpelSenf: It's documented at https://core.telegram.org/bots/api#message, too. Given how this is a main document that grammY is based on, it could make sense to change the link to that. Given how we constantly repeat information that is found on this page without naming it again, we might just as well remove the link. You decide :)

Madnex: Okay, then I'll just remove the link and leave it to the reader :P 
Don't have access to my computer right now, will update it at the end of the week 👍

roj1512: @Madnex I think you can just rest in your weekend. Thanks for the suggestion, @KnorpelSenf.

KnorpelSenf: Awesome, thanks Roj. I can offer to review.

roj1512 (commented):
LGTM. Thank you so much!

habemuscode:
https://github.com/grammyjs/website/pull/628

CikiMomogi (approved):

site/docs/plugins/session.md
```diff
@@ -326,6 +326,46 @@ If you must use the option (which is of course still possible), you should know
...
+- Only storing temporary data (or data with timeouts) in the session, and using a database for the important things that need to be migrated when a chat migrates.
+  This can then use transactions and custom logic to handle concurrent data access from the old and the new chat.
+  A lot of effort and has a performance cost, but the only truly reliable way to solve this problem.
+
+- It is theoretically possible to implement a workaround that matches both chats **without guarantee of reliability**.
+  The Telegram Bot API sends a migration update for each of the two chats once the migration was triggered (see the properties `migrate_to_chat_id` or `migrate_from_chat_id` in the [Telegram API Docs](https://core.telegram.org/bots/api#message)).
+  The issue is that there is no guarantee that these messages are sent before a new message in the supergroup appears.
+  Hence, the bot could receive a message from the new supergroup before it is aware of any migration and thus, it can not match the two chats, resulting in the aforementioned problems.
+
+- Another workaround would be to limit the bot only for supergroups with filtering (or limit only session related features to supergroups).
```
CikiMomogi, line 354:
```suggestion
- Another workaround would be to limit the bot only for supergroups with [filtering](../guide/filter-queries.md) (or limit only session related features to supergroups).
```

site/docs/id/plugins/session.md
```diff
@@ -326,6 +326,46 @@ Telegram mengirim webhook secara berurutan untuk setiap chat, oleh karena itu se
...
+  Ini adalah cara yang paling tepat dan lebih bisa diandalkan, namun membutuhkan usaha dan performa lebih.
+
+- Secara teori, kita bisa mencocokkan kedua chat tersebut.
+  Namun, **tidak ada jaminan cara ini bisa diandalkan**.
+  API Bot Telegram akan mengirim sebuah update migrasi untuk masing-masing chat ketika migrasi terjadi (lihat property `migrate_to_chat_id` atau `migrate_from_chat_id` di [dokumentasi API Telegram](https://core.telegram.org/bots/api#message)).
+  Masalahnya, tidak ada jaminan kalau update tersebut dikirim sebelum pesan baru muncul di supergroup.
+  Bot kamu bisa saja menerima pesan dari supergroup terlebih dahulu tanpa mengetahui kalau migrasi telah dilakukan.
+  Akibatnya, ia tidak bisa mencocokkan kedua chat, yang menimbulkan masalah seperti yang telah kita bahas di atas.
+
+- Solusi lainnya adalah memanfaatkan filter untuk membatasi bot supaya bisa digunakan di supergroup saja, atau bisa juga membatasi fitur yang terkait dengan session hanya untuk supergroup.
```
CikiMomogi, line 355:
```suggestion
- Solusi lainnya adalah memanfaatkan [filter](../guide/filter-queries.md) untuk membatasi bot supaya bisa digunakan di supergroup saja, atau bisa juga membatasi fitur yang terkait dengan session hanya untuk supergroup.
```

site/docs/es/plugins/session.md
```diff
@@ -326,6 +326,46 @@ Si debes usar la opción (que por supuesto sigue siendo posible), debes saber lo
...
+  Esto puede usar transacciones y lógica personalizada para manejar el acceso a datos simultáneos desde el chat antiguo y el nuevo.
+  Es mucho esfuerzo y tiene un costo de rendimiento, pero es la única forma verdaderamente confiable de resolver este problema.
+
+- En teoría, es posible implementar una solución alternativa que coincida con ambos chats **sin garantía de confiabilidad**.
+  La API de Telegram Bot envía una actualización de migración para cada uno de los dos chats una vez que se activa la migración (consulte las propiedades `migrate_to_chat_id` o `migrate_from_chat_id` en los [Documentos de la API de Telegram](https://core.telegram.org/bots/api#message)).
+  El problema es que no hay garantía de que estos mensajes se envíen antes de que aparezca un nuevo mensaje en el supergrupo.
+  Por lo tanto, el bot podría recibir un mensaje del nuevo supergrupo antes de que se dé cuenta de cualquier migración y, por lo tanto, no puede hacer coincidir los dos chats, lo que genera los problemas antes mencionados.
+
+- Otra solución alternativa sería limitar el bot solo para los supergrupos con filtrado (o limitar solo las funciones relacionadas con la sesión a los supergrupos).
```
CikiMomogi, line 354:
```suggestion
- Otra solución alternativa sería limitar el bot solo para los supergrupos con [filtrado](../guide/filter-queries.md) (o limitar solo las funciones relacionadas con la sesión a los supergrupos).
```

CikiMomogi (commented):
A little touch...

PR #564, opened 2022-12-28
Add a note and link to documentation on systemd unit files

roj1512:


site/docs/hosting/vps.md
```diff
@@ -60,6 +60,8 @@ WantedBy=multi-user.target
 ```
 
 > Replace `<start_command>` with the command you got above.
+>
+> Also note that if you Deno is installed for a different user other than root, you may need to specify it in the service section like `User=<the_user>`. For more information on the unit files, visit [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_working-with-systemd-unit-files_configuring-basic-system-settings).
```
KnorpelSenf, line 64:
```suggestion
> Also note that if Deno is installed for a different user other than root, you may need to specify it in the service section like `User=<the_user>`.
> For more information on the unit files, visit [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_working-with-systemd-unit-files_configuring-basic-system-settings).
```

KnorpelSenf (changes requested):

KnorpelSenf (approved):
LGTM, thanks!

habemuscode (approved):

PR #565, opened 2022-12-28
Fix property name

receter:
The comment says
// these are also the default values
But the value for getSessionKey is not the default.


KnorpelSenf (approved):
Whoopsie daisy, nice catch. Thanks for fixing it!

roj1512 (commented):
Please update the other languages as well.

CikiMomogi:
> Please update the other languages as well.

Done

roj1512 (approved):
LGTM. Thank you so much!

KnorpelSenf (approved):

PR #566, opened 2022-12-28
Update session.md

receter:
Proposal for a warning that Multi Sessions can not be used together with Lazy Sessions


KnorpelSenf (commented):
I feel like it makes more sense to warn about this in the lazy session docs. After all, lazy sessions haven't been introduced at this point yet. What do you think?

KnorpelSenf:
Closing due to inactivity.

PR #567, opened 2022-12-28
fix: use title case

KnorpelSenf:
Verbs are always to be capitalised in titles.


roj1512 (approved):
LGTM. Thank you so much!

PR #571, opened 2022-12-30
Fix multiple Heroku entries in Spanish comparison table

CikiMomogi:


roj1512 (approved):
LGTM. Thank you so much!

PR #572, opened 2022-12-30
fix: link to custom context type from structuring

KnorpelSenf:
We talk about how to type middleware below, but we don't mention how to annotate composers. This PQ fixes that.


roj1512 (approved):
LGTM. Thank you so much!

habemuscode (approved):

CikiMomogi (approved):

PR #577, opened 2023-01-04
Add note about specifying file extension.

mcpeblocker:
I think it would be better to add this note to inform people about the path of file. First time I looked at this I passed the path of a directory and not a file path and got `Illegal operation` error. Some people asked a solution for this and they were giving a directory path too. 
_P.S. I can't say I'm very well at English so you can change the sentences I added but including these as a note in documentation would probably benefit others and prevent getting error just because of being careless._

### DIFF
--- a/site/docs/plugins/files.md
+++ b/site/docs/plugins/files.md
@@ -93,7 +93,7 @@ bot.on([":video", ":animation"], async (ctx) => {
 </CodeGroup>
 
 You can pass a string with a file path to `download` if you don't want to create a temporary file.
-Just do `await file.download("/path/to/file")`.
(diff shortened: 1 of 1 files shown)


roj1512:
File names aren’t necessarily with an extension, and we have said `/path/to/file`, not `/path/to/dir` or something like that. Is this note really needed?

mcpeblocker:
> File names aren’t necessarily with an extension, and we have said `/path/to/file`, not `/path/to/dir` or something like that. Is this note really needed?

Oh I see your point. In that case that note wouldnʼt be necessary. Thanks 😁👍

PR #472, opened 2022-08-05
Redesign custom containers

CikiMomogi:
Enhance the tip, warning, and danger containers' color schemes. This PR may resolve #407

Before:
grammy-before

After:
gramY-after

PR #583, opened 2023-01-09
Refactor and update dependencies

CikiMomogi:
- Fix vuepress issue with dependency version range. See:
https://github.com/vuepress/vuepress-next/issues/1195.
- Add `github-pull-request` to the recommended list.
- Add `docs:preview` and `docs:deno` scripts.
- Remove css rules for custom container since vuepress has implented these rules by default. See:
https://github.com/vuepress/vuepress-next/pull/1131
- Fix github actions complaining about the deprecated node.js
- Update `package-lock.json`


site/package.json
```diff
@@ -11,15 +11,18 @@
     "docs:build": "npm run docs:vuepress && npm run docs:sitemap",
     "docs:vuepress": "vuepress-vite build docs",
     "docs:fmt": "deno fmt",
-    "docs:sitemap": "sscli -b https://grammy.dev -r docs/.vuepress/dist --format xml --no-robots --no-clean"
+    "docs:sitemap": "sscli -b https://grammy.dev -r docs/.vuepress/dist --format xml --no-robots --no-clean",
+    "docs:serve": "serve docs/.vuepress/dist",
+    "docs:deno": "deno upgrade"
```
KnorpelSenf, line 16: What is this script supposed to do?

KnorpelSenf (commented):
Can you explain the CSS changes you performed?

CikiMomogi:
> Can you explain the CSS changes you performed?

It's removing the custom CSS rules for the custom container that we added a while back (#472). We added these rules because vuepress didn't have them at the time, but they've since implemented them. I previously submitted a pull request to vuepress with the same rules that we were using in grammY. So, our custom rules are no longer necessary.

KnorpelSenf:
Pretty cool!

For future reference, it's https://github.com/vuepress/vuepress-next/pull/1131 that added these rules. This is pulled in via #583.

site/package.json
KnorpelSenf: I see, that makes sense!

However, the `docs:` prefix is meant to scope things that are about the docs themselves. Invoking a tool that runs independently of the docs doesn't really belong under the same scope IMO.

```suggestion
    "deno:upgrade": "deno upgrade"
```

How about this?

KnorpelSenf (approved):
Thank you so much!

site/package.json
```diff
@@ -11,15 +11,18 @@
...
+    "docs:sitemap": "sscli -b https://grammy.dev -r docs/.vuepress/dist --format xml --no-robots --no-clean",
+    "docs:serve": "serve docs/.vuepress/dist",
+    "deno:upgrade": "deno upgrade"
   },
   "keywords": [],
   "author": "KnorpelSenf",
   "license": "MIT",
   "devDependencies": {
-    "@vuepress/plugin-docsearch": "^2.0.0-beta.51",
+    "@vuepress/plugin-docsearch": "next",
```
KnorpelSenf, line 22: @roj1512 please confirm that you're comfortable with unpinned deps when reviewing

CikiMomogi:
https://github.com/grammyjs/website/pull/583/commits/eba86b6d982e1b410d75d8640a7141e2d5212594 - Upon discovering that `deno upgrade` was not functioning, it was replaced with `deno --version` for displaying the current version. To upgrade Deno, one can use the command `npm update` for a smooth experience.

roj1512:
Hmm

roj1512:
I badly missed this PR

site/package.json
roj1512: I'm comfortable with `@vuepress/plugin-docsearch@next`, but we might not always be fine with `vuepress-vite@next`. Revert, @CikiMomogi?

CikiMomogi:
This PQ is not urgent, so there is no need to rush or hurry. :)

site/package.json
CikiMomogi: Ok

roj1512 (approved):
LGTM. Thank you so much!

PR #586, opened 2023-01-27
Update copyright years

roj1512:


roj1512:
We were 27 days late. Happily not a month.

CikiMomogi (approved):

KnorpelSenf:
> Happily not a month.

Would anything happen if it was a month?

roj1512:
Yes. We would lose the trust of more people.

PR #600, opened 2023-02-10
fix(notifier): don’t slice the signature header

roj1512:
This should have been done since migrating octokit-webhooks-methods to v3.0.0.


KnorpelSenf (approved):
In Roj we trust

KnorpelSenf:
Why is it ready for translation? For testing purposes?

roj1512:
No. My bad.

PR #602, opened 2023-02-10
fix(notifier): correct the Telegram webhook path’s method

roj1512:


roj1512:
LAST ONE. PROMISE.

KnorpelSenf (approved):
oof

PR #605, opened 2023-02-10
fix(notifier): serious logic issues with the file proxying thing

roj1512:


KnorpelSenf (changes requested):
Please remove the patch files

roj1512:
Ik ik. I was just recovering.

notifier/main.ts
```diff
@@ -46,13 +46,17 @@ bot.chatType(["group", "supergroup"]).filter((ctx) =>
     await ctx.reply("Could not resolve PR branch.");
     return;
   }
+  const { file_path } = await ctx.api.getFile(ctx.message.document.file_id);
```
KnorpelSenf, line 49: What's wrong with using the files plugin here?

roj1512: For downloading it?

KnorpelSenf: For getting the URL

roj1512: Make a suggestion

KnorpelSenf: Did you find it?

KnorpelSenf: There's `file.getUrl`

roj1512: Oh, the URL. We can't use it for the URL.

KnorpelSenf: Why not?

roj1512: Because we don't use its URL here. We do in the `/file/:path` handler where we don't have grammY `ctx`.

KnorpelSenf: Right

KnorpelSenf (approved):
LGTM

PR #609, opened 2023-02-11
Change fonts

roj1512:


KnorpelSenf:
Is it normal that the font isn't cached? It's only applied on reload, but never on initial load, no matter how often I visit the page

CikiMomogi:
Perhaps we should consider altering the `--font-family` and  `--font-family-code` variables instead?
https://github.com/vuepress/vuepress-next/blob/4941f95da5be5a0060a916a21a62de37dd7c87b9/ecosystem/theme-default/src/client/styles/vars.scss#L76-L78

roj1512:
@KnorpelSenf Is your DevTools open when you face this problem? If yes, see the "Disable cache' checkbox.

roj1512:
@CikiMomogi Good point. But how would we change the fonts of headings?

KnorpelSenf:
No. That's how it behaves on mobile.

roj1512:
I’ll ask you to know if it reproduces later.

KnorpelSenf:
It could be that netlify disables caching for the preview URLs

roj1512:
```shell
 $ curl -is https://grammy.dev
HTTP/2 200 
age: 0
cache-control: public, max-age=0, must-revalidate
```

roj1512:
```sh
 $ curl -is https://63e78398a25c0f21912a766b--grammy.netlify.app/
HTTP/2 200 
age: 0
cache-control: public, max-age=0, must-revalidate
```

KnorpelSenf:
I think this is gorgeous, but should we maybe merge #611 into this? It seems like these two changes should be coordinated.

roj1512:
@KnorpelSenf I don’t think they’re related. They might be related by blood, but they’re separate organisms anyway.

site/docs/.vuepress/styles/_fonts.scss
```diff
@@ -0,0 +1,64 @@
...
+.site-name,
+.sidebar-heading,
+th {
+  font-family: "Red Hat Display", sans-serif;
+}
+
+pre,
+code {
+  font-family: "Source Code Pro", monospace;
+}
```
CikiMomogi, line 64:
```suggestion
kbd,
pre,
code {
  font-family: "Source Code Pro", monospace;
}
```

Or, even better, override the `--font-family-code` variable for future proofing since we use this font for all monospace.

CikiMomogi:
Can you provide us the latest preview link?

KnorpelSenf:
Yes, look at this comment to always see the latest link: https://github.com/grammyjs/website/pull/609#issuecomment-1426698750

CikiMomogi:
Hmm... Why does it fallback to sans-serif?

image

roj1512:
@CikiMomogi It is because our font-display is set to optional, and when you open your DevTools, the Disable cache option in the Network tab is most probably checked automatically, and it doesn't let the cached font to load quickly. Can you try closing the DevTools and refreshing twice?

CikiMomogi:
> Can you try closing the DevTools and refreshing twice?

Firefox: Closed Devtools, cleared data & cache, refreshed multiple times, still stuck on sans-serif fallback.

The font looks gorgeous. It is displaying correctly in Chromium, but not in Firefox.
Also, sometimes when I refresh the page or navigate between pages in Chromium without enabling DevTools, I've noticed that the correct font is being used in the body, but the fallback font Arial is used in the title of custom container.

I'd rather use the main font, even if it takes a bit longer to load. It looks so good, and I'm hoping it'll be a big draw for new visitors.

KnorpelSenf:
I believe that these problems will solve itself once we deploy it in production. If not, we can revert. Previews aren't meant to be cached, so I don't think we should really try to fix much here.

CikiMomogi (approved):

KnorpelSenf (approved):
Even after a few days of getting used it, I find the new font harder to read. It's not a lot, but it's clearly noticeable.

On the other hand, it looks a lot prettier.

I'm not against against merging because I definitely see why this is desired. I just want to make sure that we all are aware of it so that this is an informed decision :)

PR #612, opened 2023-02-11
Customize and limit selection

roj1512:
Before

image

Now

image

---

Before

image

Now

image


CikiMomogi:
Is this the latest preview? https://63e7d8178dc47b597dae386e--grammy.netlify.app/

ss

KnorpelSenf:
Yes, the comment at https://github.com/grammyjs/website/pull/612#issuecomment-1426836841 always contains the latest preview. It's edited whenever a new commit is pushed, so the link is always up to date.

site/docs/.vuepress/styles/_select.scss
```diff
@@ -0,0 +1,19 @@
+.action-button,
+.navbar-item,
+.site-name,
+.logo,
+img[alt="grammY"],
+.next,
+.prev {
+  user-select: none;
+}
```
CikiMomogi, line 9:
```suggestion
img[alt="grammY"],
.badge,
.actions,
.navbar,
.sidebar,
.footer, // footer on homescreen
footer, // footer on page
nav {
  user-select: none;
}
```

CikiMomogi: Btw, why is only the `img[alt="grammY"]` excluded, instead of all images?

CikiMomogi (changes requested):

site/docs/.vuepress/styles/_select.scss
roj1512: Good point. I was just doubtful.

roj1512:
> Is this the latest preview? [63e7d8178dc47b597dae386e--grammy.netlify.app](https://63e7d8178dc47b597dae386e--grammy.netlify.app/)

Forgot to push the latest changes.

CikiMomogi:
Is the contributor list at the bottom of the page supposed to be selectable?

roj1512:
@CikiMomogi I disabled that keeping access to the value of Last Updated. Lmk how it looks to you now.

KnorpelSenf:
I love the new highlighting, it looks a lot more shiny.

I don't really see the point why we should interdict anyone to copy other content. Does it ever get in the way? If we don't solve an immediate problem by taking away this liberty. then I'd just let people do it.

roj1512:
The places we have "interdicted" are mostly naturally already interdicted.

roj1512:
Like no one can copy an item in the dropdowns unless they do Ctrl+A.

roj1512:
Neither for the sidebar and footer buttons, unless Ctrl+A or they started the selection from another place.

roj1512:
It's also pretty uncanny to want to select the grammY logo in the header, or the button texts, don't you think?

KnorpelSenf:
Can you explain which problem you're trying to solve?

KnorpelSenf:
"it's probably not that bad because we cannot think of anything" is not really a great reason to needlessly restrict the user

roj1512:
I'm not trying to solve any problem. I have already solved a serious problem that was "restricting" the user from having a good experience.

KnorpelSenf:
I can't quite follow. What's the purpose of this PQ if it's not solving any problem?

What's that serious problem that you already have solved and that supposedly prevented good UX? I can't find any PQ about that, could you link me there?

roj1512:
I mean this selection problem solved in this PQ.

roj1512:
The problem that prevented having a good UX is the initial selectability of parts that are not supposed to be selectable, even from the user's expectations.

roj1512:
If you think the button, the grammY logo, images, or anything else that was made unselectable should be selectable, tell me why (separately, for each of them), and I'll happily revert them one by one, or even close the PQ if they were all pointlessly made selectable.

For example, for the images, I initially only made only grammY logos unselectable, but @CikiMomogi suggested to make all of the images unselectable, even those in the actual docs. I don't see a problem if they're selectable, so I can just revert that one. In short, give me concise points for each affected part.

KnorpelSenf:
No, these are completely pointless changes. There are zero reasons to pretend that we're smarter than the user and tell them what they can and cannot do. You are apparently unable to say what this is good for, so I cannot be bothered to spend time arguing about the details of anything here

roj1512:
Can I keep this issue open until I am able to say what this is good for?

KnorpelSenf (approved):

PR #614, opened 2023-02-13
Configure the Cache-Control header for fonts

roj1512:


roj1512:
Works so smooth. It feels like the font is locally installed.

image

roj1512:
Perhaps we can also add Cache-Control to other assets as well so we feel even smoother.

site/netlify.toml
```diff
@@ -0,0 +1,4 @@
+[[headers]]
+  for = "/fonts/*"
+  [headers.values]
+  cache-control = "max-age=604800"
```
KnorpelSenf, line 4:
```suggestion
  cache-control = "max-age=604800" # 1 week
```

KnorpelSenf (approved):
LGTM!

PR #615, opened 2023-02-13
feat: use Deno for sitemap generation

KnorpelSenf:
As discussed today with @roj1512 and @dcdunkan in DM.


dcdunkan (approved):
LGTM

roj1512 (approved):
LGTM.

KnorpelSenf:
@roj1512 please revert. Some pages are missing from <https://grammy.dev/sitemap.xml>. I would like to investigate the differences before fixing it.

Try running the build locally before/after merging and diff the files.

PR #616, opened 2023-02-13
Trigger build on workflow_dispatch

roj1512:
This is for when we want to manually trigger a build.


KnorpelSenf (approved):
Do I understand it correctly that we are able to manually trigger CI with this?

PR #617, opened 2023-02-15
Use Cache-Control for images

roj1512:


site/package.json
```diff
@@ -23,5 +23,8 @@
     "deno-bin": ">=1.25.2",
     "serve": "^14.1.2",
     "vuepress-vite": "2.0.0-beta.60"
+  },
+  "dependencies": {
+    "glob": "^8.1.0"
```
KnorpelSenf, line 28: Why is this necessary?

roj1512: Mistake

KnorpelSenf (approved):
LGTM, thanks! Anything left to do here?

roj1512:
Merged before the build passing hence #619

PR #624, opened 2023-02-19
RU: fix format of files by removing TODO prefixes

evermake:


evermake:
@KnorpelSenf , is `Translation Guard / check-labels` check necessary?

KnorpelSenf:
Yes, it tells us if all languages are completely translated. You can bypass it by marking a pull request as not translatable.

IMO it's also acceptable to temporarily disable this check on your translation branch as long as we remember to introduce it again before completing the last translations of the docs.

KnorpelSenf (approved):
LGTM

PR #625, opened 2023-02-19
Fix dates and links of the last Bot API update

evermake:
Seems like Telegram updated the link and date of the last update, or we initially made it wrong...


KnorpelSenf (approved):
It seems like Telegram changed the date afterwards. Thanks for fixing it!

roj1512 (approved):
LGTM. Thank you so much!

PR #633, opened 2023-02-26
Implement Webhook Utility

roziscoding:
Successor to https://github.com/grammyjs/website/pull/523

Layout preview:

image
image
image


roziscoding:
I created a new dedicated page for this component and added a link to it in the "Resources" menu in the navbar.
That's a suggestion and I'm completely fine if we decide to change it

KnorpelSenf:
1. Pretty awesome stuff! 🎉
2. What's the loading spinner at the top doing? It never completes for me 🤔
3. It would be cool to remove the button to load the bot info. This can happen automatically as soon as the token is inserted and correct.
4. Does the component actually store the token in localStorage? It's gone upon reload.
5. We should link to this website's source code in the promise that we don't send the token anywhere.
6. Are you planning on adding a light mode?
7. This should probably be in its own category under Resources, perhaps called Tools. http://fqs.deno.dev can go in there, too, as soon as it's done. /cc @dcdunkan
8. Additionally, we should add this component to every tutorial on the page that tells users to set a webhook.

roziscoding:
> What's the loading spinner at the top doing? It never completes for me 🤔
It's loading the bot's picture. I don't think it's able to handle a bot not having a picture. I'll fix that.

> It would be cool to remove the button to load the bot info. This can happen automatically as soon as the token is inserted and correct.
I can't really validate the token without calling the API, and I'm not sure if it's a good idea to do that on every typed character, but if you think it's fine, I can implement it no problem

> Does the component actually store the token in localStorage? It's gone upon reload.
No, not really. I forgot about that part. I'll add it ASAP

> We should link to this website's source code in the promise that says we don't send the token anywhere.
👍🏻 

> Are you planning on adding a light mode?
The component framework already has one built-in, but I couldn't figure out how to detect that the user changed the overall page theme

> This should probably be in its own category under Resources, perhaps called Tools. http://fqs.deno.dev/ can go in there, too, as soon as it's done. /cc @dcdunkan
I agree.

> Additionally, we should add this component to every tutorial on the page that tells users to set a webhook.
Should we do this on a new PR after this one gets merged?

KnorpelSenf:
> it's a good idea to do that on every typed character, but if you think it's fine

I do think it's fine. Note that tokens are usually pasted, so most often there will only be a single request. Even if there are a few hundred requests, it shouldn't be much of an issue.



> I couldn't figure out how to detect that the user changed the overall page theme

There's https://github.com/vuepress/vuepress-next/blob/d283ffe3ef0668bfea54e6d973066695f46f13c0/ecosystem/theme-default/src/client/composables/useDarkMode.ts but it doesn't seem to be available to the outside. We can check the DOM manually, but that's ugly. Perhaps @Loskir or @CikiMomogi know this?



> Should we do this on a new PR after this one gets merged?

That depends on whether or not we even want to keep it on a dedicated page. Note that #635 was closed. @roj1512 what do you think?

dcdunkan:
> I can't really validate the token without calling the API, and I'm not sure if it's a good idea to do that on every typed character, but if you think it's fine, I can implement it no problem

Why not use a regex for quickly validating that the bot token is at least valid. Then if it is valid, then request?

KnorpelSenf:
Because the format of the tokens is unspecified, so we could only base this regex on our own assumptions. The correct way to validate a token is via `getMe`.

Also, trying to be smart by avoiding certain calls is nothing but a performance optimisation. As long as we don't actually know of any problems related to performance, there's no need to spend time on it.

Also also, rate limits are not a problem. They're very high for `getMe`, so no ordinary website visitor could ever trigger it, even not by typing the token by hand many times. What's more: the rate limits enforced on checking bot token validity (=failing calls) are enforced per IP address AFAIK. This means that malicious website visitors can only harm themselves but never us. (I'm ignoring the fact that malicious actors can spam the endpoint a million times from a single bash command so there's no need for them to even load a website.)

roj1512 (commented):
Should this also be in a separate repository?

KnorpelSenf:
This is a grey area. It is not exactly markdown text (it is not just docs) but then again it's just a component that this website uses, and that no other website uses. If we were to move it to a different repo, we would consequently also have to extract things like https://github.com/grammyjs/website/blob/main/site/docs/.vuepress/plugins/better-line-breaks.ts but this leads to a lot of added complexity. I am not sure what we gain by this.

I don't think that this component can be treated the same way as the filter query explorer.

What we can do is to only include it on the tutorial pages, and drop the Resources > Tools entry.

KnorpelSenf:
It still is a separate VuePress plugin that can be used independently from this repo, so we could publish it on npm and then consume it. I just don't wanna do it to avoid the maintenance overhead.

roziscoding:
> Because the format of the tokens is unspecified, so we could only base this regex on our own assumptions. The correct way to validate a token is via `getMe`.
> 
> Also, trying to be smart by avoiding certain calls is nothing but a performance optimisation. As long as we don't actually know of any problems related to performance, there's no need to spend time on it.
> 
> Also also, rate limits are not a problem. They're very high for `getMe`, so no ordinary website visitor could ever trigger it, even not by typing the token by hand many times. What's more: the rate limits enforced on checking bot token validity (=failing calls) are enforced per IP address AFAIK. This means that malicious website visitors can only harm themselves but never us. (I'm ignoring the fact that malicious actors can spam the endpoint a million times from a single bash command so there's no need for them to even load a website.)

Okay. I'll implement it then.

roziscoding:
> Should this also be in a separate repository?

I don't think it's necessary since this shouldn't really be needed outside the docs.

Also, publishing it as a library would imply setting up the build system and everything.

Also also, I'd feel the need to make it more generic in terms of appearance and I really don't know / wanna know how to do that with the component library I'm using for this.

This repo contains a website, and this PR adds a component of that website. I really think they fit together nicely.

roziscoding:
> There's https://github.com/vuepress/vuepress-next/blob/d283ffe3ef0668bfea54e6d973066695f46f13c0/ecosystem/theme-default/src/client/composables/useDarkMode.ts but it doesn't seem to be available to the outside.

You can usually import hooks from `@vuepress/client`, but I don't know about that one.
Ill check it out as soon as I get to my computer

roj1512:
> It still is a separate VuePress plugin that can be used independently from this repo, so we could publish it on npm and then consume it. I just don't wanna do it to avoid the maintenance overhead.

Right, but it is still something that cannot be interacted with in the front end unlike the webhook utils component.

roj1512:
But anyway, let’s continue this way. Will this component be in every page that guides the user to set webhook?

roj1512 (changes requested):
Please rename `en-US.ts` to `en_US.ts`, `es-ES.ts` to `es_ES.ts`, and so on.

roziscoding:
> Will this component be in every page that guides the user to set webhook?

That's the ideia, yeah.
But I'd rather merge this first, then create another PR to add it to those pages

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
+import { Translation } from "./types";
+
+export const lang: Translation = {
+  tokenCard: {
+    disclaimer: {
```
roj1512, line 5: This is not a disclaimer. Change this key to `warning` or sth similar.

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+      title: "About your token",
```
roj1512, line 6:
```suggestion
      title: "Disclaimer",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+      content:
+        "We never store your bot's token in the backend. It is stored in your browser's localStorage. Do not go around giving your token to random websites, no matter how nicely they ask!",
```
roj1512, line 8:
```suggestion
        "We don’t take your bot token to the back end. It will only be stored in your browser’s local storage. Don’t provide your bot token to places you don’t trust.",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+  botCard: {
+    canJoinGroups: "Can join groups",
+    canReadGroupMessages: "Can read all group messages",
+    inlineQueries: "Supports inline queries",
+    tabs: {
+      webhookInfo: "Webhook Info",
+      manageWebhook: "Manage Webhook",
+    },
+    buttons: {
+      changeToken: "Change token",
```
roj1512, line 29:
```suggestion
      changeToken: "Change Token",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+    },
+  },
+  manageWebhook: {
+    fields: {
+      secret: {
+        label: "Webhook Secret",
```
roj1512, line 35:
```suggestion
        label: "Webhook secret",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+          required: "Please specify a URL",
+        },
+      },
+      dropPending: {
+        label: "Drop pending updates",
+      },
+    },
+    buttons: {
+      setWebhook: "Set webhook",
+      deleteWebhook: "Delete webhook",
```
roj1512, line 53:
```suggestion
      deleteWebhook: "Delete Webhook",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+          protocol: "Telegram only allows HTTPs",
+          required: "Please specify a URL",
+        },
+      },
+      dropPending: {
+        label: "Drop pending updates",
+      },
+    },
+    buttons: {
+      setWebhook: "Set webhook",
```
roj1512, line 52:
```suggestion
      setWebhook: "Set Webhook",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+      deleteWebhook: "Delete webhook",
+    },
+  },
+  webhookInfo: {
+    empty: "No webhook set",
```
roj1512, line 57:
```suggestion
    empty: "No webhook is set",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+    pendingUpdates: (count: number) =>
+      `${count} pending update${count !== 1 ? "s" : ""}`,
+    lastErrorDate: (formattedDate: string) => `Last error (${formattedDate})`,
```
roj1512, line 60:
```suggestion
    lastErrorDate: (formattedDate: string) => `Last error: ${formattedDate}`,
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+        label: "Webhook Secret",
+        placeholder: "Secret value telegram sends to your bot on every update",
+      },
+      url: {
+        label: "Webhook URL",
+        placeholder: "Full URL to your webhook",
+        errorMessages: {
+          invalid: "Invalid URL",
+          protocol: "Telegram only allows HTTPs",
+          required: "Please specify a URL",
```
roj1512, line 44:
```suggestion
          required: "Specify a URL",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+      secret: {
+        label: "Webhook Secret",
+        placeholder: "Secret value telegram sends to your bot on every update",
+      },
+      url: {
+        label: "Webhook URL",
+        placeholder: "Full URL to your webhook",
+        errorMessages: {
+          invalid: "Invalid URL",
+          protocol: "Telegram only allows HTTPs",
```
roj1512, line 43:
```suggestion
          protocol: "The main Bot API server only allows HTTPS.",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+    fields: {
+      secret: {
+        label: "Webhook Secret",
+        placeholder: "Secret value telegram sends to your bot on every update",
+      },
+      url: {
+        label: "Webhook URL",
+        placeholder: "Full URL to your webhook",
+        errorMessages: {
+          invalid: "Invalid URL",
```
roj1512, line 42:
```suggestion
          invalid: "Invalid URL specified.",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+    },
+    buttons: {
+      changeToken: "Change token",
+    },
+  },
+  manageWebhook: {
+    fields: {
+      secret: {
+        label: "Webhook Secret",
+        placeholder: "Secret value telegram sends to your bot on every update",
```
roj1512, line 36:
```suggestion
        placeholder: "Secret value that will be sent with every update",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
roziscoding: jsyk, button text is full caps always

roj1512: I already know. It must be consistent in the source code as well :)

roj1512 (commented):
Can you make the preview work somehow?

roziscoding:
> Can you make the preview work somehow?

On it. I saw it was failing after I requested your review.

roj1512 (commented):
The cursor for the disabled checkboxes should be either `default` or `not-allowed`, not `pointer`.

roj1512 (commented):
Do you think it should adapt light mode better? Or this dark background is intended?

image

roj1512 (commented):
Do some `overflow:hidden; text-overflow: ellipses` so longer webhook URLs appear better.

image

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+export const lang: Translation = {
+  tokenCard: {
+    disclaimer: {
+      title: "Disclaimer",
+      content:
+        "We don’t take your bot token to the back end. It will only be stored in your browser’s local storage. Don’t provide your bot token to places you don’t trust.",
+    },
+    fields: {
+      token: {
+        label: "Bot Token",
```
roj1512, line 12:
```suggestion
        label: "Bot token",
```

site/docs/.vuepress/components/webhook-util/translations/en_US.ts
```diff
@@ -0,0 +1,72 @@
...
+        placeholder: "The one which @BotFather gave you.",
+      },
+    },
+    buttons: {
+      loadBotInfo: "Load Bot Info",
```
roj1512, line 17: Is this string used?

roziscoding: Not anymore, no.

roj1512: Remove it then.

KnorpelSenf:
I have some more feedback. All of these ideas are optional and won't block from merging. You're free to decide if you want to do them here or in a subsequent PQ or not at all.

1. The three checkboxes look like we could change the settings. It is not intuitive that all they do is to provide information. Either restyle, or add a tooltip, or add a link to BotFather, or a combination of the above.
2. The text field loses focus when typing.
3. The error message "Network request for 'getMe' failed!" should not be printed when the text field is cleared.
4. The error message "Network request for 'getMe' failed!" should perhaps be changed to something more informative.
5. The text "The one which @BotFather gave you." should include a link to BotFather.
6. The back arrow should perhaps be displayed next to the bot's avatar. That's where I'd expect backwards nav.
7. The component should not flicker when pressing refresh.

KnorpelSenf:
I have some more feedback. All of these ideas are optional and won't block from merging. You're free to decide if you want to do them here or in a subsequent PQ or not at all.

1. The three checkboxes look like we could change the settings. It is not intuitive that all they do is to provide information. Either restyle, or add a tooltip, or add a link to BotFather, or a combination of the above.
2. The text field loses focus when typing.
3. The error message "Network request for 'getMe' failed!" should not be printed when the text field is cleared.
4. The error message "Network request for 'getMe' failed!" should perhaps be changed to something more informative.
5. The text "The one which `@BotFather` gave you." should include a link to BotFather.
6. The back arrow should perhaps be displayed next to the bot's avatar. That's where I'd expect backwards nav.
7. The component should not flicker when pressing refresh.

roj1512:
Closing due to inactivity and grammY tools.

KnorpelSenf:
Wouldn't it still be awesome to have a webhook utility component that can be included in the respective guides? IMO it is much preferable to write docs that go like

_you gotta set your token, have a tool for that, look:_ [TOOL]

rather than

_you gotta set your token, you can do this via curl or your browser or this other website over there which we created, it is called tools.grammy.dev and it is very good but no matter what you do you always have to leave this site and your train of thought_

roziscoding:
I agree it's better to have something inline, but I'll open another PQ with the grammy tools version of the component, since it doesn't depend on a UI framework and the design should work better in the docs

KnorpelSenf:
Awesome!

roj1512:
Thanks for proving what I assumed as a YES @roziscoding.

PR #634, opened 2023-02-28
Use a more accurate term

KnorpelSenf:
Uses a more accurate term


CikiMomogi (approved):

CikiMomogi:
Should we change this line too?
https://github.com/grammyjs/website/blob/c5237d7c724ad5450727d60e3f69f326c424fc47/site/docs/guide/deployment-types.md?plain=1#L360

KnorpelSenf:
No. In fact, it seems like I am misusing the term data race, as most people only refer to a case where two threads access the same memory location as a data race. JS is single-threaded, so by definition there are no data races. This PQ is wrong.

If you want to apply the unrelated changes to ID anyway, kindly do so in a separate PQ, as I will close this.

PR #635, opened 2023-03-01
feat: Filter Query Browser

### LINKED ISSUE #158: feat: Interactive filter query explorer
Paste your filter query, and get visualised what it means, and what kinds of updates would be caught. Not sure how much effort this is. Ideally, we don't have to manually keep this in sync with the code.

We might have to drop this if it is too much work. After all, filter queries are not _that_ hard to understand, so it might not justify the effort.

dcdunkan:
Closes #158. (Its not really interactive tho)

Notes:

- I'd appreciate if someone can implement the search component, if that is needed. As @roziscoding have offered to help, I guess he is ready to implement one 😄?
- @KnorpelSenf should this be ignored algolia? If so, can you update the configuration?
- I'm not sure how the generated files should be ignored by git, if they have to be.
- Consider helping improving the docs part please.
- To the translators, you can see the translations/en.ts file to know how the translations should go. And they need to be imported in translations/mod.ts. That's it.
- The prefixDocs in en.ts translation is only a demonstration. I think its nice. Can be removed if its not really useful. If it is useful, consider adding more to the list. @KnorpelSenf, if you could choose a right set of filter queries that needs additional explanation, that'll be useful.
- This is not actually a plugin. Can be moved to another directory (tools?) if needed.

Feedbacks welcome.

### DIFF
--- a/site/docs/.vuepress/configs/en.ts
+++ b/site/docs/.vuepress/configs/en.ts
@@ -334,8 +334,8 @@ export const localeEn: LocaleConfig<DefaultThemeLocaleData> = {
               {
                 text: "Filter Query Browser",
                 link: "/resources/tools/filter-queries.html",
-              }
-            ]
+              },
+            ],
           },
         ],
       },
--- a/site/docs/.vuepress/configs/en.ts
+++ b/site/docs/.vuepress/configs/en.ts
@@ -328,6 +328,15 @@ export const localeEn: LocaleConfig<DefaultThemeLocaleData> = {
               },
             ],
           },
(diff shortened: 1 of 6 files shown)


KnorpelSenf:
Please be transparent about why this was closed. DMs aren't known publicly.

dcdunkan:
Oh, sorry. I did it again.

dcdunkan:
Closed because we decided to move it to a separate webpage because its a tool, and not a part of the actual documentation. This repository should be only for documentation.

PR #636, opened 2023-03-02
Remove notifier

roj1512:
As it was moved to an independent repository.


KnorpelSenf (approved):
It's a good idea to refocus this repo. Thanks.

PR #640, opened 2023-03-02
(Ukrainian) Translate main guide page

Nazar-Ant:


roj1512:
@Nazar-Ant If you want the checks to pass, add the not translatable label.

Nazar-Ant:
@roj1512 Yes, I already guessed. Thanks for the tip.

PR #649, opened 2023-03-06
Globes for language switching

KnorpelSenf:
How else would I know how to switch back the language?

### DIFF
--- a/site/docs/.vuepress/configs/en.ts
+++ b/site/docs/.vuepress/configs/en.ts
@@ -12,7 +12,7 @@ export const siteEn: SiteLocaleConfig = {
 
 export const localeEn: LocaleConfig<DefaultThemeLocaleData> = {
   "/": {
-    selectLanguageText: "Languages",
+    selectLanguageText: "🌎",
     selectLanguageName: "English",
     editLinkText: "Edit this page on GitHub",
     notFound: [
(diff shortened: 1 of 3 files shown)


roj1512 (approved):
LGTM. Thank you so much!

roj1512:
We could use a single globe for all languages if the displayed regions are a problem.

KnorpelSenf:
> I wonder why globes? In my opinion, they are not suitable because they only depict three major regions of the Earth: America, Africa and Asia. Why not use a flag emojis, for example?

Which continents are you missing? Antarctica?

KnorpelSenf:
> We could use a single globe for all languages if the displayed regions are a problem.

Dropping the rotation part effectively puts a single continent into focus, which isn't exactly better. Using 🌐 looks ugly.

Nazar-Ant:
Have you considered using a [similar icon](https://fontawesome.com/icons/language?s=solid&f=classic)?

roj1512:
I mean, I’m not a designer but I think it would need a lot of thinking to choose a suitable external icon.

KnorpelSenf:
> Have you considered using a [similar icon](https://fontawesome.com/icons/language?s=solid&f=classic)?

No. So far, we have only considered Unicode characters for simplicity. Does the globe bother you that much?

PR #660, opened 2023-03-08
Revert "feat: use Deno for sitemap generation"

KnorpelSenf:
This reverts commit 5fdb4d81bbd13cbc9d7cf7c53b36bc7b31d9b39f from #615.

This is necessary because the Deno tool was rushed, and it is nowhere close to being on par with the old solution. While it is still desirable to move to Deno tooling eventually, we should first make sure that the sitemap CLI on Deno creates a byte-identical file with the `sscli` tool that is reintroduced in this PQ. If any differences are accepted, they should be backed up by a good reason.

The Google Search Console reports a massive increase in 404 links since a few days after #615 was merged.

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -11,7 +11,7 @@
     "docs:build": "npm run docs:vuepress && npm run docs:sitemap",
     "docs:vuepress": "node --max-old-space-size=8000 ./node_modules/.bin/vuepress-vite build docs",
     "docs:fmt": "deno fmt",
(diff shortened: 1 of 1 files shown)


dcdunkan:
Were you able to look into the issue why a lot of pages were missing in the generated sitemap.xml? I tried generating one locally and it had all of it.

dcdunkan (approved):
LGTM

KnorpelSenf:
> Were you able to look into the issue why a lot of pages were missing in the generated sitemap.xml? I tried generating one locally and it had all of it.

No, I did not investigate this yet.

KnorpelSenf:
This PQ was announced in https://github.com/grammyjs/website/pull/615#issuecomment-1428744173 already.

roj1512:
Why haven’t we merged this?

KnorpelSenf:
Because it's just 8 minutes old.

PR #662, opened 2023-03-09
Migrate back to the Deno-powered sitemap generation

KnorpelSenf:
This is good now, I fixed up a lot of things and it produces the same `sitemap.xml` file now. The only difference is that it has a better sorting than `sscli`.

### DIFF
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "grammy-website",
   "version": "1.0.0",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


roj1512 (approved):
LGTM.

CikiMomogi:
I don't know much about XML, but I'm wondering if the double slash `//` in the URL is valid?
```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
    <url>
        <loc>https://grammy.dev//404.html</loc>
        <lastmod>2023-03-04T17:10:07.523Z</lastmod>
    </url>
    ...
    <url>
        <loc>https://grammy.dev//resources/faq.html</loc>
        <lastmod>2023-03-04T17:09:55.590Z</lastmod>
    </url>
</urlset>
```

CikiMomogi:
> It is not. Where do you see this?

Oh nvm. It's good now. I forgot to run `docs:sitemap` :P

KnorpelSenf:
> > It is not. Where do you see this?
> 
> Oh nvm. It's good now. I forgot to run `docs:sitemap` :P

Okay awesome! This and #660 were supposed to fix exactly that issue 😅

PR #663, opened 2023-03-09
Update deployment-types.md

Bekoff257:


KnorpelSenf:
Which question does this answer? Is it not obvious that a bot communicating with Telegram is a Telegram bot?

Using a much longer term will inevitably make the entire ASCII drawing wider, so it may no longer fit as well on all screens. We try to keep these illustrations as narrow as possible.

PR #671, opened 2023-03-10
Bot tokens everywhere

KnorpelSenf:
While bot tokens technically are authentication tokens, everyone just calls them bot tokens. We even do this in all code snippets.

We should use the more common term. It is also the simpler one.

I added another sentence that makes clear that bot tokens serve the same purpose as authentication tokens.


roj1512 (approved):
LGTM. Thank you so much!

habemuscode (approved):

fadzikri:
Bruh, i forgot. That's package name.

fadzikri (approved):
ok!

PR #673, opened 2023-03-11
Fix import in chat-members typescript example

v9gel:


roj1512:
Since grammY 1.10.0, types are exported from the core package, and installing `@grammyjs/types` is no longer required.

PR #685, opened 2023-03-12
[Indonesian] Add explanation about this 34-second video

CikiMomogi:


fadzikri (approved):
Nice improvitation. 🚀

KnorpelSenf (approved):
Excellent! 🎉🎉🎉

This is the exact reason why we need you guys. No translation software could do something like this. Outstanding idea! 😍

roj1512 (approved):
Looks liquidable to me. Thank you <3

PR #734, opened 2023-03-22
Fix broken anchor link

roj1512:


KnorpelSenf (approved):
Tested the anchor, it works

PR #740, opened 2023-03-23
Use /types everywhere

### LINKED ISSUE #701: Use /types in examples
The `/types` feature should always be used to import types for the Bot API. However, we sometimes don't even do this on the code examples: https://grammy.dev/guide/context.html#via-inheritance

This needs to be fixed everywhere.

KnorpelSenf:
Closes #701.

### DIFF
--- a/site/docs/es/guide/context.md
+++ b/site/docs/es/guide/context.md
@@ -432,7 +432,10 @@ bot.start();
 
 ```ts
 import { Bot, Context } from "https://deno.land/x/grammy/mod.ts";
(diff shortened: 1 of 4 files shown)


roj1512 (approved):
LGTM. Thank you so much!

CikiMomogi:
Should this be included as well?
https://github.com/grammyjs/website/blob/556a890b508fe9d3d94ae4cbe75b37727109c0ec/site/docs/guide/api.md?plain=1#L79

KnorpelSenf:
No, it was intentionally omitted. We still want to link to the original source of the types. We explain right after this that they are exported from grammY, and how they can be used.

PR #741, opened 2023-03-23
Implement language popup

roj1512:


site/docs/.vuepress/components/LanguagePopup.vue
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const popup = reactive({
+  enabled: localStorage.getItem("disable_language_bar") == null,
+});
+
+const language = computed(() =>
+  popup.enabled && navigator.language in languages
+    ? [navigator.language, languages[navigator.language]]
+    : null
```
KnorpelSenf, line 17: I think we need to check all `navigator.languages` here

roj1512: What if it included all the supported languages?

KnorpelSenf: Take the first.

KnorpelSenf (changes requested):
There is a layout shift on mobile when the popup is closed by pressing the X.

Download the images and switch back and forth. Note that the Quickstart section moves down.

Before:
localhost_8080_
After:
localhost_8080_ (1)

KnorpelSenf (approved):
LGTM, TYSM!

KnorpelSenf:
Do you want @wojpawlik to review this PQ or view the merged result?

site/docs/README.md
```diff
@@ -93,3 +93,7 @@ Works! :tada:
 (Last highlight: Translated Bot Descriptions)
 
 </div>
+
+<ClientOnly>
+  <LanguagePopup />
+</ClientOnly>
```
CikiMomogi, line 99: Is it possible to use the [layout slots](https://v2.vuepress.vuejs.org/reference/default-theme/extending.html#layout-slots) instead of manually adding components to each page?

CikiMomogi: Why not?

roj1512:
@KnorpelSenf @wojpawlik prefers ready tables.

PR #760, opened 2023-03-25
Translate #758 and #759

roj1512:
Sync #758 and #759

### DIFF
--- a/site/docs/id/README.md
+++ b/site/docs/id/README.md
@@ -86,7 +86,7 @@ Berhasil! :tada:
 ---
 
 <ClientOnly>
-  <ThankYou :s="['Terima kasih ', 'Someone', ' telah menjadi kontributor grammY.']" />
+  <ThankYou :s="['Terima kasih, ', 'Someone', ', telah menjadi kontributor grammY.', ', telah menciptakan grammY.']" />
 </ClientOnly>
 
 <div style="font-size: 0.75rem; display: flex; justify-content: center;">
--- a/site/docs/id/README.md
+++ b/site/docs/id/README.md
@@ -86,7 +86,12 @@ Berhasil! :tada:
 ---
 
 <ClientOnly>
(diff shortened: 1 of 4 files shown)


CikiMomogi (approved):

site/docs/id/README.md
```diff
@@ -86,7 +86,7 @@ Berhasil! :tada:
 ---
 
 <ClientOnly>
-  <ThankYou :s="['Terima kasih ', 'Someone', ' telah menjadi kontributor grammY.']" />
+  <ThankYou :s="['Terima kasih, ', 'Someone', ', telah menjadi kontributor grammY.', ', telah menciptakan grammY.']" />
```
fadzikri, line 89: Wait... i don't understand here. But, ok. I will see after it deployed.

fadzikri (approved):

site/docs/id/README.md
CikiMomogi: This PQ contains changes from #758 and #759.

#758: When 'KnorpelSenf' (the founder of grammY) appears on the list of contributors, we now use the sentence "Thank you, KnorpelSenf, for _**creating**_ grammY" instead of "Thank you, KnorpelSenf, for being a _**contributor**_ to grammY".

#759 is a fix for the punctuation in the EN doc. Since the punctuation rules are the same in both English and Indonesian, we applied these rules to this PQ and added several commas in certain places.

CikiMomogi: Btw @roj1512, is it safe to change the 'Someone' to '$name'?

roj1512: Yes

fadzikri: @CikiMomogi i see. i understand now, i just confuse in https://github.com/grammyjs/website/pull/758 beetwen that sentences. ok... 🚀

KnorpelSenf: I hope it's possible to share the prefix between both translations. This may work for the current languages, but not necessarily for all future ones. We may need a pair of string triples here.

habemuscode (approved):

PR #774, opened 2023-05-04
Add a warning about using render.com for polling in the comparisons section.

ahmeddots:
I went through a lot of trouble setting up the bot and the database with render.com, only to be greeted by an error that affects bots deployed with said problem.

The issue is that render does not stop the other instance before deploying a new one, which makes it so that you're polling with two instances, and this causes the following error:

`GrammyError: Call to 'getUpdates' failed! (409: Conflict: terminated by other getUpdates request; make sure that only one bot instance is running)`

I believe it is good to warn users about using render for hosting bots that use polling.

### DIFF
--- a/site/docs/hosting/comparison.md
+++ b/site/docs/hosting/comparison.md
@@ -45,6 +45,12 @@ The main thing to know about them is that on serverless infrastructures you are 
 | DigitalOcean Apps      | $5         |                                                                                                            |                                                                                         | ❓      | ❓                           | ❓  | Not tested                                                                                       |
 | Fastly Compute@Edge    |            |                                                                                                            |                                                                                         | ❓      | ❓                           | ❓  |                                                                                                  |
 
+::: tip About render.com
+
+Render implements zero downtime deploys, meaning you'll always have one instance running. This will cause issues with polling, since only one instance can poll the telegram servers for updates.
+
+:::
+
 ### VPS
 
 A virtual private server is a virtual machine that you have full control over.


KnorpelSenf (commented):
It seems like https://github.com/grammyjs/website/pull/774/files#diff-5abe1fca9d392457ba58ace5ade18f9d5f53d796c4bf77357ae5a3219a41f6f1R30 already explains exactly this, so your addition is redundant. @roj1512 what do you think?

ahmeddots:
> It seems like https://github.com/grammyjs/website/pull/774/files#diff-5abe1fca9d392457ba58ace5ade18f9d5f53d796c4bf77357ae5a3219a41f6f1R30 already explains exactly this, so your addition is redundant. @roj1512 what do you think?

`"...on serverless infrastructures you are required to use webhooks."`

That is incorrect, in the context of your comparison table. I've noticed that just now. I run  my bot with long polling, on Fly.io, which you've marked as **serverless**, but my bot works just fine.

There needs to be a differentiation between platforms/services that allow you to upload functions such as DigitalOcean Functions and Cloudflare Workers. These are **true serverless**, where you upload functions, which run when hit or called, then stop after they finish (typically a few seconds). Which is why they cannot be used for polling.

And the other kind of serverless (I hesitate to call them that, myself) that includes Render.com, Railway, and Fly.io. These do not let you directly manage or access the machine similar to a VPS, but allocate you resources and lets your codebase running without stop.

So, no. The issue I encountered isn't with serverless, and that line you've marked does not explain the problem. It is incorrect as well.

KnorpelSenf:
I can't quite follow. It says "all entries below require webhooks" basically. Now you want to add "this one entry in particular requires webhooks".

Which of the hosting providers in the serverless table should be used with long polling in your opinion?

ahmeddots:
All the entries do **not** require webhooks.

Fly.io works with long polling. I'm literally running a bot on it with long polling. And (I assume) DigitalOcean and Heroku will also be able to run a bot with long polling. Because they offer non-serverless options.

The issue here is with your classification of services/platforms, and I believe the page is misusing the word **_serverless_**.

There are essentially three kinds of hosting options:
1. Serverless (write functions which only run when called, and timeout after a few seconds)
3. VPS (have full access to a virtual server)
4. PaaS (deploy code without worrying about infrastructure management)

The page is mixing up Serverless and PaaS. I'd be happy to open a pull request for that.

roj1512:
@ahmeddots See #777.

ahmeddots:
Good work clarifying that, @roj1512. Though I believe the warning of this PR should still be included.

roj1512:
I personally think it’s very specific, but my thought won’t be taken for granted if you all agree that we need this.

ahmeddots:
@roj1512, I'm just hoping to save someone from wasting a lot of time deploying it to Render like I did.

Also, I noticed we're from the same city. Very cool.

roj1512:
@ahmeddots We’ll have to wait to see what @KnorpelSenf thinks.

> Also, I noticed we're from the same city. Very cool.

Woah! Nice to meet you.

ahmeddots:
@roj1512, of course.

Same here! I don't see a lot of us in OSS.

roj1512:
Well, you now see me :D

Also feel free to join [grammY’s group](https://t.me/grammyjs) if you wanted to talk more.

KnorpelSenf:
I still believe that this is redundant information, but how about we try to be better at what we're already saying? What I mean by this is something like https://github.com/grammyjs/website/pull/777/files#r1208379709.

ahmeddots:
> I still believe that this is redundant information, but how about we try to be better at what we're already saying? What I mean by this is something like [#777 (comment)](https://github.com/grammyjs/website/pull/777#discussion_r1208379709).

Of course that works, too. I suggested that a re-categorization was due in the previous comments. I'll be closing the issue now.

PR #779, opened 2023-05-26
Set font-display to swap

roj1512:


KnorpelSenf (approved):
I don't like it, but I like the current solution even less. Maybe wait for @Loskir to confirm?

roj1512:
Okay.

PR #783, opened 2023-05-27
Configure font with variables

roj1512:


Loskir (approved):
LGTM but we should check whether `<kbd>` looks as intended

roj1512:
Seems fine.

image

PR #788, opened 2023-06-09
Disambiguate type exports

KnorpelSenf:
I don't want to make people think that they need to install a new package. Some folks refuse to read on so they miss the code examples.


site/docs/guide/api.md
```diff
@@ -77,7 +77,7 @@ grammY allows you to specify objects consistently across the API, and makes sure
 
 grammY ships with complete type coverage of the Bot API.
 The [`@grammyjs/types`](https://github.com/grammyjs/types) repository contains the type definitions that grammY uses internally.
-These type definitions are also exported so you can use them in your own code.
+These type definitions are also directly exported from the `grammy` core package so you can use them in your own code.
```
roj1512, line 80:
```suggestion
These type definitions are also directly exported from the core `grammy` package so you can use them in your own code.
```

KnorpelSenf: I think both versions work but they express different things, no?

roj1512: My interpretation:

- _`grammy` core package_ is like _`grammy` is not the only core package_.
- _core `grammy` package_ is like _the core package aka `grammy`_.

What’s yours?

KnorpelSenf: I think we have the same definition. My intent was to say _core package_ and then specify the name of the package, so it's the _`grammy` core package_. But I think people will know that we mean either way so we can call it the _core `grammy` package_, too, even though it sounds to me like there were several packages called `grammy` and we specify that we mean the core one.

KnorpelSenf: I'll leave it up to you, feel free to accept your own suggestion if you want to go ahead

roj1512 (approved):

LWJerri (approved):

fadzikri (approved):

habemuscode (approved):

PR #789, opened 2023-06-10
Add unused but explanatory async

KnorpelSenf:
In response to https://github.com/grammyjs/grammY/issues/431

@dcdunkan are there any more such places? We have dozens or even hundreds of code examples with async handlers, but the rule is that anything that can be misunderstood will be misunderstood sooner or later. It's probably best to always add async, except for lambdas without block such as `(ctx) => ctx.reply("42")`.


site/docs/es/guide/middleware.md
```diff
@@ -84,7 +84,7 @@ Este stack de funciones es la _middleware stack_.
 (ctx, next) => ...    |
 ```
 
-Volviendo a nuestro ejemplo anterior, ahora sabemos por qué `bot.on(":photo")` ni siquiera se comprobó: el middleware en `bot.on(":text", (ctx) => { ... })` ya se encargó de la actualización, y no llamó a `next`.
```
dcdunkan, line 87: @KnorpelSenf I think here the async is unnecessary (and in the translations). What do you think?

site/docs/advanced/transformers.md
```diff
@@ -51,7 +51,7 @@ Calls via context objects of concurrently running middleware are left unaffected
 As soon as the respective middleware completes, the transformer function is discarded.
 
 ```ts
-bot.on("message", (ctx) => {
+bot.on("message", async (ctx) => {
```
KnorpelSenf, line 54: We should call next here

site/docs/es/plugins/session.md
```diff
@@ -669,7 +669,7 @@ En la práctica, en lugar de tener los datos de la sesión disponibles en `ctx.s
 
 ```ts
 // Sesiones por defecto (sesiones estrictas)
-bot.command("settings", (ctx) => {
+bot.command("settings", async (ctx) => {
```
KnorpelSenf, line 672: Why is the highlighting broken here, is there no parse?

dcdunkan: Not sure, but the file looks correct locally. Seems like just an highlighting issue on Github's side.

site/docs/advanced/transformers.md
dcdunkan: In this pull request, or should we do this in another?

site/docs/es/guide/middleware.md
KnorpelSenf: I actually agree

site/docs/advanced/transformers.md
KnorpelSenf: In a subsequent PQ. It's kinda unrelated.

roj1512 (approved):
I am also approving these changes on behalf of Dunkan since we are in each other’s souls.

KnorpelSenf:
Ready to merge then :)

roj1512:
I will take that as it is not translatable.

PR #794, opened 2023-06-15
we all know they do

KnorpelSenf:
so we tell them to just not


site/docs/id/guide/deployment-types.md
```diff
@@ -236,6 +236,8 @@ app.use(webhookCallback(bot, "oak"));
 :::
 ::::
 
+> Kamu tidak perlu memanggil `bot.start()` ketika menggunakan webhooks.
```
fadzikri, line 239:
```suggestion
> Kamu tidak perlu memanggil `bot.start()` ketika menggunakan webhook.
```

Do we still need `s` in bahasa?

fadzikri (changes requested):
@CikiMomogi what's your opinion about this?

LWJerri (approved):

site/docs/id/guide/deployment-types.md
CikiMomogi: I am not sure though. What do you think?

KnorpelSenf: @fadzikri?

CikiMomogi: I think both of them is okay. But for the shake of consistency, we should determine which one should be used. This one is safe to be merged.

KnorpelSenf:
Another thing that is blocked by #822

PR #810, opened 2023-06-19
Configuring Markdownlint

Nazar-Ant:


.github/workflows/check.yml
```diff
@@ -19,6 +19,21 @@ jobs:
...
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 'latest'
+
+      - name: Check Linting
+        working-directory: ./site
+        run: |
+          npm i -D markdownlint-cli2
```
KnorpelSenf, line 34: Why not `npm install`?

site/.markdownlint.json
```diff
@@ -0,0 +1,11 @@
+{
+  "line-length": false,
+  "no-inline-html": false,
```
KnorpelSenf, line 3: Do we need inline HTML?

KnorpelSenf (changes requested):

site/.markdownlint.json
Nazar-Ant: You are using some Vue components and divs on the home page. You can't get rid of them, so I just disabled this rule.

KnorpelSenf: Those are exceptions, I think we generally don't want HTML (since we now no longer need code groups with HTML). So it's better to add linting exception comments to these places, and have the rule enabled. What do you think?

.github/workflows/check.yml
roj1512: Why not `npm ci`?

Nazar-Ant: With npm ci, it is not possible to install a specific dependency. So we can only install all dependencies, which is not optimal. However, the guarantee that npm ci will install only the version specified in the package.json is a good enough reason to use npm ci. If installing extra packages is not a problem, I will use npm ci.

KnorpelSenf: It isn't, go ahead. The important part is not to take more time than a build, but we're never gonna need that much time anyway

Nazar-Ant: That is, use npm ci?

roj1512: Yes.

habemuscode (approved):

Nazar-Ant:
I think I did everything I wanted to do. Currently, linting is not working because of the `deno-bin` removal, so I'll wait until it's back and then move this PR to "ready for review".

Nazar-Ant:
@KnorpelSenf @roj1512 This PR is synchronized with the main branch. Everything is working.

site/.markdownlint.jsonc
```diff
@@ -0,0 +1,8 @@
+{
+  "line-length": false,
+  "no-duplicate-heading": false,
```
KnorpelSenf, line 3: Where does this fail?

site/.markdownlint.jsonc
```diff
@@ -0,0 +1,8 @@
...
+  "no-trailing-punctuation": {
+    "punctuation": ".,;:。，；：" // everything except exclamation points
```
KnorpelSenf, line 5:
```suggestion
    "punctuation": ".,;:。，；：" // everything except exclamation points and question marks
```

KnorpelSenf (changes requested):
Should we maybe disable `link-fragments` entirely? It looks like the better-anchors plugin breaks this (which is fine).

@dcdunkan does the link-checker make sure that all internal links to anchors are correct?

site/.markdownlint.jsonc
KnorpelSenf: Oh, alright

Nazar-Ant:
> Should we maybe disable `link-fragments` entirely? It looks like the better-anchors plugin breaks this (which is fine).

If your tool checks anchors, we can indeed disable this rule.

site/.markdownlint.jsonc
Nazar-Ant: Screenshot from 2023-07-07 16-16-52

KnorpelSenf: I was referring to `no-duplicate-heading`

KnorpelSenf: But why does `no-duplicate-heading` fail for long lines?

KnorpelSenf: Or were you talking about different things? I understand that in the Hosting/VPS section we have repeated headers, but that makes sense because we describe the same set of steps twice in two different contexts.

All the other reported cases can be fixed, noß

Nazar-Ant: No, this rule does not complain about long headlines. It's just that you gave an example of code for two rules, and I responded accordingly.

KnorpelSenf: Ah well yes, GH always renders a few contextual lines of code before, but I added my comment to line 3 and I meant to refer to `no-duplicate-heading` only :)

Nazar-Ant: In `hosting /vps.md` we can insert a comment at the beginning of the page to disable the `MD024/no-duplicate-heading/no-duplicate-header` rule globally on the page. On other pages, you should rephrase the headers:

- `plugins/keyboard.md`. Complaint about the repeated header `Responding to Clicks`, which occurs in the context of the inline and custom keyboard. At the same time, headers such as `Building ...`, `Sending ...` and `Removing ...` are used with the keyboard in mind: `Building an Inline Keyboard' and `Building a Custom Keyboard', for example. You can do the same with a problematic header: `Responding to Clicks on the Inline Keyboard` and `Responding to Clicks on the Custom Keyboard`.
- `hosting/heroku.md`. Complaint about the repeated header `Creating bot.ts`, which occurs in the context of webhooks and a long polling. We can do the same thing here as I suggest for `plugins/keyboard.md`, i.e. create the headers `Creating bot.ts for Webhooks` and `Creating bot.ts for a Long Polling`.
- There are still a few pages (`id/plugins/conversations.md`, `es/guide/filter-queries.md`) where translators have somehow duplicated certain headings, although they are all unique in the original.

KnorpelSenf: I'll fix these things in a subsequent PQ targeting this branch.

KnorpelSenf:
Blocked by https://github.com/grammyjs/website/pull/810#discussion_r1257043490

Nazar-Ant:
@KnorpelSenf Check the result on the website. If you accept this, then I should add indentation to the code blocks that follow the last item in the list if the code relates to it.

KnorpelSenf:
The indentation looks good to me. It isn't applied everywhere yet, correct?

KnorpelSenf:
Nevermind, it seems like the only cases where this is missing are exactly as you has described, only after lists. Please go ahead then :)

KnorpelSenf:
Anything else before we can move this out of draft state?

Nazar-Ant:
No, everything is ready.

Nazar-Ant:
@KnorpelSenf I remembered that we hadn't yet figured out whether to keep my plugin (https://github.com/grammyjs/website/pull/810#pullrequestreview-1518834431) and what to do with the `no-duplicate-heading` rule (https://github.com/grammyjs/website/pull/810#discussion_r1256982641).

KnorpelSenf:
Let's keep the plugin, but apply all of the changes you suggested in https://github.com/grammyjs/website/pull/810#discussion_r1256982641. Could you take care of that?

dcdunkan:
> @dcdunkan does the link-checker make sure that all internal links to anchors are correct?

Sorry, I missed this.

To make clear, you're asking, whether we check if the anchor exists in links like `./file.md#anchor` , right? Then yes.

KnorpelSenf:
Yes, that's what I was asking. Nice! That gives us some more safety.

Nazar-Ant:
@KnorpelSenf Do I have to remove the plugin then?

site/.markdownlint.jsonc
Nazar-Ant: Okay, I'll do it.

KnorpelSenf:
Just to sum it up and check if we are all on the same page, this is our reasoning. Please correct me if anything is wrong.

1. Vitepress messes up the anchors if certain characters are used in headings.
2. This can be fixed by introducing markdown linting to interdict the problematic characters.
3. Some languages need these characters in their words.
4. @Nazar-Ant wrote the better-anchors plugin which replaces some characters for the anchor links, hence allowing us to use them in headings.
5. This extended support means we had to adjust the list of characters that the linting would forbid.
6. @dcdunkan wrote the link checker which analyses the entire website post-build and detects if Vitepress messes up headings as explained in 1.
7. We can therefore disable the linting for headings.
8. We can therefore remove the better-anchors plugin.

Did I miss anything?

Nazar-Ant:
That's right. So, I remove the plugin.

KnorpelSenf:
Yep!

dcdunkan:
> 6. @dcdunkan wrote the link checker which analyses the entire website post-build and detects if Vitepress messes up headings as explained in 1.
> ...
> Did I miss anything?

Yes, one thing, sorry that I didn't mention it here before. Link checker currently does its thing before building the website, not after. But it uses the almost-same configuration for the markdown parser as Vitepress. I'm not entirely sure if Vitepress does anything else to anchors or headings after parsing the markdown; I hope not and as far as I checked it doesn't. But if it does, then the results produced by the link checker won't be accurate. I'm not sure what to do with this information, but I thought everyone should know this before making any changes.

KnorpelSenf:
Ah. Thanks interesting. But it checks external links with anchors for correctness, right? Does it also check the anchors after build? Or does it compare the links against headers?

Nazar-Ant:
> Ah. Thanks interesting. But it checks external links with anchors for correctness, right?
> Does it also check the internal anchors against the build? Or does it compare the links against headers?

Vitepress simply replaces certain characters and that's it. But actually, this is probably done by MarkdownIt (a library for parsing .md files). So we just need to do the same thing that has already been implemented in link-checker.

Nazar-Ant:
It may be worth looking at the source code of Vitepress and MarkdownIt to find the exact rules for building links and replacing characters.

KnorpelSenf:
Does it ever replace characters in links differently from how it replaces characters in headings?

Nazar-Ant:
No

KnorpelSenf:
I guess then we should be fine

.github/workflows/check.yml
```diff
@@ -19,6 +19,21 @@ jobs:
...
         run: deno fmt --check
 
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 'latest'
```
roj1512, line 29:
```suggestion
          node-version: latest
```

site/.markdownlint.jsonc
roj1512: Was this issue resolved?

Nazar-Ant: Work in progress https://github.com/grammyjs/website/pull/908

KnorpelSenf (approved):

roj1512:
Add the not translatable label, @Nazar-Ant?

Nazar-Ant:
@roj1512 Well, in general, yes, there is nothing to translate. At the beginning of the creation, I didn't know if anything would need to be translated (in fact, some things did), so I didn't add the label. However, now it's really worth adding this label so that the github doesn't get angry at checks.

roj1512 (approved):
Thank you!

PR #814, opened 2023-06-20
Better anchors

Nazar-Ant:
I have written a plugin that searches for problematic characters and words in links and replaces them with similar ones that are contained in the links generated by VuePress (VitePress generates the same links, so this plugin will be useful during migration). This plugin will help to get rid of errors due to violation of the `MD051/link-fragments` rule, which are typical not only for the Ukrainian language.


KnorpelSenf (commented):
What effects does this have?

- Editor tooling? Auto-complete suggests headers in markdown. Will they be correct?
- Search? Algolia links to sections. Will this index update accordingly after crawling the site again?
- Navigation? The page overview lets us scroll to sections. Will this keep on working correctly?

Nazar-Ant:
> Editor tooling? Auto-complete suggests headers in markdown. Will they be correct?

When you type `#`, the code editor offers you that auto-generated anchor, which does not always match the anchor used by VuePress. This plugin allows you to leave an auto-generated anchor in the code that Markdownlint won't complain about, but when the site is launched, this anchor will be slightly modified to match the VuePress anchor.

> Search? Algolia links to sections. Will this index update accordingly after crawling the site again?

Everything should be fine, because the plugin modifies the anchors during the build, so only the correct links will remain in the html.

> Navigation? The page overview lets us scroll to sections. Will this keep on working correctly?

Yes, the links from the table of concepts (sidebar) coincide with the links generated by the plugin.

Nazar-Ant:
Although I absolutely lack the knowledge to write an effective plugin. Perhaps it would be better to completely replace the function that VuePress uses to create the link, but I don't know if that's possible.

KnorpelSenf:
Thanks for the explanations!

If anything, we contribute back to Vitepress :)

KnorpelSenf (approved):
Thank you so much!

PR #819, opened 2023-06-21
Replace deno-bin with me

roj1512:


KnorpelSenf (approved):
Just for the record I want to have stated that this will make it harder for people to contribute so it is to be expected that someone will be annoyed sooner or later

PR #820, opened 2023-06-21
Update package.json

roj1512:


roj1512:
Lmk if there’s more useless fields. I would personally just keep the dependencies.

KnorpelSenf (approved):

site/package.json
```diff
@@ -1,9 +1,7 @@
 {
-  "name": "grammy-website",
-  "version": "1.0.0",
-  "description": "Main documentation site for grammY",
-  "main": "index.js",
   "private": true,
+  "license": "MIT",
+  "author": "KnorpelSenf",
```
KnorpelSenf, line 4: You can remove this, it's a bit tilted anyway, I'm by far not the only author here

KnorpelSenf:
> just keep the dependencies

I think it's good to have the scripts and `private` and `license`

PR #824, opened 2023-06-22
Migrate to VitePress

CikiMomogi:
:red_circle:  LIVE Demo : https://github.com/grammyjs/website/pull/824#issuecomment-1604200758

#### Vitepress:

> home-new

<details>
  <summary>Click to see MOREEEEE</summary>
  
#### Vuepress:

> home-old
---

### Docs
---
---

#### Desktop - Homescreen
---
---

#### Mobile - Homescreen
---
---

#### Desktop - Docs

---

---

#### Mobile - Docs
---

---

</details>

## Major changes
I made some major changes in this PQ. Please fill the checkbox if you are okay with it :
- [x] Convert all images to .webp to reduce the size file. But, the original files (jpeg/png) is not deleted.
- [ ] CleanUrls is enabled. Need some configuration on hosting side. But, if you are deploying the site in Netlify, their cleanurl settings should be enabled by default.
- [x] All reference link that ended with .md extensions is omitted as the [recommendation from Vitepress](https://vitepress.dev/guide/routing#linking-between-pages)
- [x] Use local search instead of Algolia. I feel like this local search is more accurate and has better UI/UX, especially the detailed result feature
- [x] Dark mode is enabled by default

## TO DO
I can't fix this. Help wanted! :
- [x] Migrating all vuepress plugins
- [ ] Fix chunk size limit warning during build process. The temporary workaround is by enlarge the chunk size limit ([3fad343c](https://github.com/grammyjs/website/pull/824/commits/3fad343c2483c7bf1d28f269fa5ddeb430761dd2))
- [x] Fix hydration error at homepage. Log: https://pastebin.com/ePVWuEh0
- [x] Fix warning `The resource at “https://cm-grammy.netlify.app/assets/inter-roman-latin.2ed14f66.woff2” preloaded with link preload was not used within a few seconds. Make sure all attributes of the preload tag are set correctly. cm-grammy.netlify.app`


dcdunkan:
This looks so cool.

> Convert all images to .webp to reduce the size file

If you're doing that please make sure you have also uploaded their JPEG versions as well; Instant-View doesn't support WEBP format yet. So, we're doing it like this. See https://github.com/grammyjs/website/pull/824.

> All reference link that ended with .md extensions is omitted as the [recommendation from Vitepress](https://vitepress.dev/guide/routing#linking-between-pages)

The link checker probably needs updating.

roj1512 (commented):
Quotes aren’t smart.

CikiMomogi:
> If you're doing that please make sure you have also uploaded their JPEG versions as well; Instant-View doesn't support WEBP format yet. So, we're doing it like this. See #427.

Noted! https://github.com/grammyjs/website/pull/824/commits/496b103fc9bd127503e6686d2d772c55654fec78

> Quotes aren’t smart.

What?!

roj1512:
'' should be ‘’
...

KnorpelSenf:
We need to enable the vitepress equivalent for https://github.com/grammyjs/website/blob/9b0197419c2acb817db6ab22fa95ce6141d0815b/site/docs/.vuepress/config.ts#L48

Nazar-Ant:
> All unsupported syntax highlighting format in block code (procfile, log, ftl, asciiart, etc) is converted to txt to suppress the warning during build process.

You should use `text`, not `txt`, since the former is currently used.

CikiMomogi:
> We need to enable the vitepress equivalent for
> 
> https://github.com/grammyjs/website/blob/9b0197419c2acb817db6ab22fa95ce6141d0815b/site/docs/.vuepress/config.ts#L48

Please check https://cm-grammy.netlify.app/typo. Is it good now?

KnorpelSenf:
Yes, excellent!

roj1512 (changes requested):
Please fix the conflicts.

roj1512 (commented):
If you replace the txt with text, will it warn you?

CikiMomogi:
> If you replace the txt with text, will it warn you?

Yes. A lot of warn.

CikiMomogi:
Changelog:
- Syntax highlighting: All languages are now supported. No more warnings!
- Plugins: All VuePress plugins have been installed successfully.

CikiMomogi:
> Todo: Fix chunk size limit warning during build process. The temporary workaround is by enlarge the chunk size limit ([3fad343c](https://github.com/grammyjs/website/pull/824/commits/3fad343c2483c7bf1d28f269fa5ddeb430761dd2))
> ```
> (!) Some chunks are larger than 500 kBs after minification. Consider:
> - Using dynamic import() to code-split the application
> - Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
> - Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
> ```

Found the culprit. It's caused by [localsearch](https://vitepress.dev/reference/default-theme-search#local-search) indexing size (> 500 kBs).

localsearch index

CikiMomogi:
I have compared the performance of Algolia search and minisearch (the search we are using now) on PC and mobile devices.

The result is that minisearch is faster to display search results even though the index size exceeds the recommended limit. Minisearch has a split-second delay for initial initiation (probably due to reading the index), but once loaded and used it runs smoothly.

While Algolia has no delay during initiation but when used the delay is quite noticeable and repeats in every search because it has to fetch the index from the internet.

Due to the small size of the index file, no significant performance impact, and other advantages such as detailed results, I think it's worth sticking with this search. We can ignore the warning by increasing the file size limit as we currently do.

KnorpelSenf:
image
This is being displayed incorrectly

.gitignore
```diff
@@ -1,2 +1,3 @@
 /.idea/
 .DS_Store
+/site/.vitepress
```
roj1512, line 3: Seriously?

KnorpelSenf: What is the point of that?

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "tabWidth": 2,
+  "singleQuote": false,
+  "trailingComma": "all",
+  "semi": true
+}
```
KnorpelSenf, line 6: Do we no longer use Deno tooling now?

.vscode/settings.json
```diff
@@ -2,8 +2,7 @@
   "deno.config": "./deno.jsonc",
   "deno.enable": true,
   "deno.enablePaths": [
-    // everything except .vuepress
-    "./site/README.md",
```
KnorpelSenf, line 6: Why this exception?

.gitignore
CikiMomogi: what? why? I am asking the same question here. lol

CikiMomogi: Let me fix the merge conflict first

CikiMomogi:
> image This is being displayed incorrectly

Elaborate please. Working fine for me

.vscode/settings.json
CikiMomogi: No idea. It's been there since the beginning

.prettierrc
CikiMomogi: It used to help me to format files automatically. Should we also remove prettier from the recommended list of vscode extensions (and other rarely used extensions)?

.vscode/settings.json
dcdunkan: > Why this exception?

Because .vuepress/ contains vuepress plugins and configurations? (Node)

KnorpelSenf: I mean why do you no longer want to enable Deno for the file `./site/README.md`? So far, it has worked well to have it enabled.

CikiMomogi: The file was written twice by mistake. Have a look at it again, the README file is still there.

.gitignore
CikiMomogi: Oh. Turns out it was due to the default cache directory location being there. I have moved the location to `./site/docs/.vitepress` (notice the `docs`) -> https://github.com/grammyjs/website/pull/824/commits/5825a097d9a2a64973f7580e9d2be07bd26beafb

KnorpelSenf:
Now there are previous/next links between plugin pages, which doesn't make sense and is undesired. Also, the links from the last page in the guide to the first page of the advanced section and back are missing. Can these things be fixed?

CikiMomogi:
> Now there are previous/next links between plugin pages, which doesn't make sense and is undesired.
I thought it was a good idea to make it easier for users to navigate to other plugins.

> Also, the links from the last page in the guide to the first page of the advanced section and back are missing.
It is intentional. Are you sure you want me to connect these two sections?

KnorpelSenf:
> > Now there are previous/next links between plugin pages, which doesn't make sense and is undesired.
> 
> I thought it was a good idea to make it easier for users to navigate to other plugins.

That makes it seem like there's an order on the plugins. The plugins are not related to each other, so they shouldn't be linked.

> > Also, the links from the last page in the guide to the first page of the advanced section and back are missing.
> 
> It is intentional. Are you sure you want me to connect these two sections?

I assumed that the two sections were currently linked, but this isn't the case, so your changes are good. I'm not sure which is better, but for now, let's just leave things as they are.

KnorpelSenf (commented):
Thanks!

KnorpelSenf:
> > image This is being displayed incorrectly
> 
> Elaborate please. Working fine for me

When I open the preview, it suggests to switch to Ukrainian. I am not in Ukraine nor do I speak the language. The logic that determines whether to show this popup seems to be flawed.

KnorpelSenf (changes requested):
External links are missing the symbol to indicate that they open in new tabs.

KnorpelSenf:
1. The tag lines on the front page should end with a full stop (.)
2. The back button in the search does not close the search
3. Can we make the emoji for scalable be ⚡ or 🏆 or 🥇 instead of 📈?

roj1512:
@KnorpelSenf What are the language settings of your browser?

KnorpelSenf:
German

roj1512:
I need to see screenshot evidence.

roj1512:
Send me the output of `navigator.languages` from your browser’s console.

KnorpelSenf:
On mobile. Will reproduce on desktop when I return home.

KnorpelSenf:
The filter query page seems to be missing.

CikiMomogi:
> The filter query page seems to be missing.

https://649ed7c41a12d11065bc537d--grammy.netlify.app/guide/filter-queries


> External links are missing the symbol to indicate that they open in new tabs.

Coming very very soon: https://github.com/vuejs/vitepress/pull/2501


>     2. The back button in the search does not close the search

You mean this back button, right? The button closes the search just fine for me.

search

KnorpelSenf:
> I need to see screenshot evidence.

Fixed

KnorpelSenf:
> > The filter query page seems to be missing.
> 
> https://649ed7c41a12d11065bc537d--grammy.netlify.app/guide/filter-queries

Only the link in the menu was missing, but I can no longer reproduce this

> > External links are missing the symbol to indicate that they open in new tabs.
> 
> Coming very very soon: [vuejs/vitepress#2501](https://github.com/vuejs/vitepress/pull/2501)

Nice!

> > ```
> > 2. The back button in the search does not close the search
> > ```
> 
> You mean this back button, right? The button closes the search just fine for me.
> 
> search

It only works correctly after typing a search query.

CikiMomogi:
> It only works correctly after typing a search query.

Right. I see it now.

CikiMomogi:
Anyway, @KnorpelSenf . I've been meaning to ask this for a long time. Do you think it's a good idea to fork the vitepress repository and customize it to what we want. That way we have our own vitepress package.

The reason is that I feel it's very complicated to modify the core vue components. One of the reasons why I closed PQ https://github.com/grammyjs/website/pull/507 is there are too many works to be done. By creating our own version of package, we can flexibly change it and more easily synchronize the changes with upstream.

KnorpelSenf:
I think you're way more knowledgeable with Vitepress than me, so you should be the one to decide this. I won't stand in the way, but I would like to point out a few (maybe obvious) things before we fork because it would be bad to say “I could've told you” after we forked.

- We don't simply adjust it a bit. It effectively means to create a new project in a new repo which again requires maintenance and so on. We should be aware that this requires time and effort and someone needs to step forward as responsible for it.
- An alternative to forking is to contribute things back upstream to Vitepress. For example, perhaps there's a way to change Vitepress so that it's easier to customize, or perhaps some of the changes we want for us will benefit other users of Vitepress, too?

If we have someone to start a fork and we also want to perform changes that are unwelcome upstream, then we just need green light from @roj1512 and then we can go ahead with this.

roj1512:
@CikiMomogi Go with it.

CikiMomogi:
Thank you for your permission.

CikiMomogi:
> 2. The back button in the search does not close the search

:heavy_check_mark:  https://github.com/vuejs/vitepress/pull/2566

roj1512 (commented):
Can we use images instead of emojis with the cards in the main page?

CikiMomogi:
Yes, we can. svg, png, jpeg...

roj1512:
Can you take the PNGs of the current emojis from https://github.com/korenskoy/emoji-data-ios and use those?

CikiMomogi:
ss

roj1512:
This is better. Can you make them a little smaller as well?

roj1512 (changes requested):
Can we add this to the custom CSS?

```scss
.VPButton {
	user-select: none;
}
```

site/docs/README.md
```diff
@@ -1,43 +1,52 @@
...
+      text: Get Started
+      link: ./guide/getting-started
+    - theme: alt
+      text: Introduction
+      link: ./guide/introduction
+
 features:
-  - title: Easy to Use
+  - icon: <img class="VPImage" src="/icons/beach-animation.webp" alt="beach animation" width="32" height="32">
+    title: Easy to Use
```
roj1512, line 29:
```suggestion
    title: Easy-to-use
```

roj1512 (approved):

KnorpelSenf (approved):

PR #825, opened 2023-06-22
Update Deno Deploy Pricing

CetinSert:
image


KnorpelSenf (changes requested):
Thanks! Are you able to spot the same numbers on the pages of the other languages, and update them as well? Then we can merge immediately

CetinSert:
Cloudflare Workers, and Vercel are the same.
I don't know the others.

KnorpelSenf:
I think I wasn't clear. The website is available in 5 languages and we have a team of translators who are busy syncing things up. This means that the same number has to be fixed on the other 4 pages.

KnorpelSenf:
Never mind, I think the other languages are very outdated for some reason. Just leave things as they are, we're gonna take care of it.

Thanks!

roj1512 (approved):
Thanks!

KnorpelSenf (approved):

PR #829, opened 2023-06-26
Fix links

dcdunkan:


KnorpelSenf:
`/uk/` is correct?

roj1512 (approved):
Thanks!

dcdunkan:
> `/uk/` is correct?

Apparently, yes.

PR #834, opened 2023-06-27
Simplify import

PonomareVlad:
As we [tested](https://github.com/grammyjs/grammY/pull/441) earlier, the `"browser"` target in the `"exports"` field is already supported by **Vercel Edge** and **Cloudflare Workers**

I suggest shortening the optional import from the `web` path to get more generic examples

[Example source](https://github.com/PonomareVlad/grammYExportsTest/blob/cloudflare/bot.mjs#L1)


roj1512 (approved):
Thank you!

roj1512:
Hmm seems like there is nothing to do here. The imports were already replaced to deno.land/x in main.

KnorpelSenf:
Perhaps we should do the same for Vercel?

roj1512:
Convert to deno.land/x import or just grammy?

KnorpelSenf:
I was thinking of using `grammy` instead of `grammy/web` but now that you ask about it, we might actually migrate the guides to the new deno-based service by vercel, right?

roj1512:
What is the new deno-based service by vercel?

KnorpelSenf:
https://vercel.com/docs/functions/edge-functions is based on https://deno.com/subhosting

roj1512:
Oh that makes Sense

roj1512:
I’ll leave that decision to our VP of Serverless, @PonomareVlad.

PonomareVlad:
> Hmm seems like there is nothing to do here. The imports were already replaced to deno.land/x in main.

@roj1512 [There is to do](https://github.com/grammyjs/website/pull/916)

roj1512:
Thanks!

roj1512:
What are your thoughts on https://github.com/grammyjs/website/pull/834#issuecomment-1762788608, @PonomareVlad?

PonomareVlad:
> What are your thoughts on https://github.com/grammyjs/website/pull/834#issuecomment-1762788608, @PonomareVlad?

[There is no any Deno-based service in Vercel](https://t.me/grammyjs/175428?thread=175413) 🌚

PonomareVlad:
Except [this community builder for Serverless Functions ](https://github.com/vercel-community/deno)

roj1512:
What about their Edge Runtime? https://github.com/grammyjs/website/pull/834#issuecomment-1762797800

PonomareVlad:
> What about their Edge Runtime? https://github.com/grammyjs/website/pull/834#issuecomment-1762797800

Earlier I think about extending Vercel guide for Edge Runtime case, but we need to test which grammY plugins still incompatible with it and notify about this before announcing compatibility 🌚

roj1512:
So you mean the `grammy` import is better for now, right?

PonomareVlad:
> So you mean the `grammy` import is better for now, right?

Is similar to `grammy/web`

roj1512:
So it’ll be fine for Vercel, right?

PonomareVlad:
> So it’ll be fine for Vercel, right?

Including Vercel, right

roj1512:
Can you create a PQ for that?

PonomareVlad:
Yes, but it depends on:

> > What about their Edge Runtime? https://github.com/grammyjs/website/pull/834#issuecomment-1762797800
> 
> Earlier I think about extending Vercel guide for Edge Runtime case, but we need to test which grammY plugins still incompatible with it and notify about this before announcing compatibility 🌚
> 
> [For example](https://github.com/grammyjs/conversations/pull/82#issuecomment-1751989236)

roj1512:
I meant for the Node.js example, not their edge runtime.

PonomareVlad:
> I meant for the Node.js example, not their edge runtime.

In docs or in repo with examples ?

PR #836, opened 2023-06-28
feat: ensure snippet is fully working on its own

jebreuer:
I just had to dig through some stuff and test until I got it working. Just want to save others the hassle. ;)


KnorpelSenf:
Using menus inside conversations has so many pitfalls that we no longer recommend doing this. Have a look at the current version of the page: https://grammy.dev/plugins/conversations#working-with-plugins

PR #828, opened 2023-06-26
Unrecommend using menus and conversations together

KnorpelSenf:
https://t.me/grammyjs/145782

PR #842, opened 2023-07-01
Fix Missing Chinese Translation in Code

grammyz:
ref #828


roj1512 (approved):
Why did I miss this so bad

PR #843, opened 2023-07-01
[ID] Change Title Pages in Plugins

fadzikri:
cc : @CikiMomogi

What is your opinion about this?


fadzikri:
Hey @CikiMomogi, would you agree this? that word "Tanpa" changed to "Nir". I mean like "Nirlaba" or something from "Endless" which meaning "Nirakhir" instead of "Tanpa Akhir".

But, I think it's also weird.

CikiMomogi:
> Hey @CikiMomogi, would you agree this? that word "Tanpa" changed to "Nir". I mean like "Nirlaba" or something from "Endless" which meaning "Nirakhir" instead of "Tanpa Akhir".
> 
> But, I think it's also weird.

You know, I was pretty stuck on finding the right translation of that "stateless question" phrase. But you nailed it. Adding "nir" is a brilliant idea. The meaning is more or less similar to "non-status question", but using the word "nir" sounds more suitable.

CikiMomogi:
Can you remove the changes that are not related to this PQ? Like changing a plugin list that belongs to another PQ.

fadzikri:
Sure.

PR #857, opened 2023-07-04
Fix spacing

CikiMomogi:


CikiMomogi:
The spacing between these two headers is too close https://grammy.dev/resources/comparison#summary

roj1512 (approved):
In CikiMomogi we trust.

PR #858, opened 2023-07-04
Convert and optimize the grammY logo and other assets

CikiMomogi:


KnorpelSenf:
Why add JPEG? They're several times larger than the PNGs

CikiMomogi:
No specific reason actually, just to provide more compatibility. I thought since we use JPEG for Telegram Instant-view, why not keep it in this folder in case it's needed someday.

KnorpelSenf:
We only do this for the keyboard screenshots because that is their original format. Everything else is PNG.

CikiMomogi:
Yes, we are not using the JPEG version of `Y` and `grammY`. It is just for convenient in case someone need them in JPEG format.

CikiMomogi:
Btw if you are wondering why `filter-query-size.png` is more smaller than `filter-query-size.jpg`, it is because we are adding the white background to the JPEG for better visibility. There reason I was choosing the JPEG format instead of PNG for the new file is because the former is more smaller in size.

KnorpelSenf:
> Yes, we are not using the JPEG version of `Y` and `grammY`.

Yes, we do use them.

> It is just for convenient in case someone need them in JPEG format.

Why? We already have PNG versions, which are better. What is the point of converting the assets to other formats that no one needs?

Maybe this is not clear.

1. Our own logos are designed as SVG, and are converted to PNG for IV and to WEBP for the web. (We can consider serving SVG files.)
2. Screenshots are taken as PNG, and are used as PNG on IV and converted to WEBP for the web.
3. The examples files from Telegram are downloaded as JPEG (ew, stupid) and are used as JPEG on IV (same servers) and then converted to WEBP for the web.

There is no reason to convert more files to JPEG.

CikiMomogi:
> Why? We already have PNG versions, which are better. What is the point of converting the assets to other formats that no one needs?
> There is no reason to convert more files to JPEG.

It's useless for us. I'm 100% agree. JPEG is there for people other than us that may need them (eventhough the precentage is very very small). Since it has no benefical to us, either delete them or save them, I have no problem at all.

> 2. Screenshots are taken as PNG, and are used as PNG on IV and converted to WEBP for the web.

Correction according to our current IV rules:

2.1 Screenshots are taken as PNG, and converted to WEBP (if it has a smaller size) for the web, and are used as JPEG on IV. Since PNG is not used at all, we can safely delete them.

2.2 Screenshots are taken as PNG, and are used as PNG on both IV and Web. (If the converted WEBP has a bigger size)

KnorpelSenf:
Yes, we should delete all files that are 100 % useless. We only keep the original versions of all files, as well as the stuff we actually serve (either on IV or on the web).

dcdunkan:
I thought we were going to convert all JPEGs to PNG. But right now, it seems like we have JPEGs for WEBPs and PNGs for SVGs.

CikiMomogi:
> I thought we were going to convert all JPEGs to PNG. But right now, it seems like we have JPEGs for WEBPs and PNGs for SVGs.

I don't mind though. Do you approve this, @roj1512 @KnorpelSenf ?

KnorpelSenf:
I see no reason to use JPEG for any graphic on the website We do not serve photos.

roj1512:
@KnorpelSenf Do you have a final thought on this or I can take it?

KnorpelSenf:
I think I have said everything that I think of this, feel free to take over

roj1512 (approved):

Issue #855, opened 2023-07-04
Algolia crawling fails


KnorpelSenf:
`There was a blocking error and the crawler has been paused. Please resolve it before moving forward. Without intervention, the crawl will be discarded and retried on the next schedule (2 attempts remaining).`

![IMG_20230704_193726_301](https://github.com/grammyjs/website/assets/12952387/e8b77a44-574e-4990-99cf-f1c9cdba08c7)

Related: https://www.algolia.com/doc/tools/crawler/apis/configuration/

CikiMomogi:
Can you show me your crawler configuration?

KnorpelSenf:
We didn't configure it, it's Algolia that hosts it for us. They used to have their configuration stored in https://github.com/algolia/docsearch-configs/blob/master/configs/grammy.json and https://github.com/algolia/docsearch-configs/blob/master/deployed-configs/g/grammy.js but then the migrated their infrastructure to … something very different (I never really tried to understand it) and now it's stored somewhere on their servers.

I can try to dig through the dashboards in some time and see if I find anything that's related to this issue. Either way, the config has never been changed.

CikiMomogi:
> https://github.com/algolia/docsearch-configs/blob/master/configs/grammy.json

Ah, I think I found the culprit. They are looking for vuepress class which is doesn't exist. My cm-grammy.netlify.dev just got approval for their crawling program. I will try to do some experiment tonight and give you the fixed config asap (hopefully).

KnorpelSenf:
Don't bother. I got time to look into this. I found https://docsearch.algolia.com/docs/templates/#vitepress-template and will fix it now.

CikiMomogi:
Nice. Pretty outdated though. For example, lvl0 should select the active navlink in sidebar. Still better than nothing.

KnorpelSenf:
Yep it works pretty poorly, I'm still investigating

KnorpelSenf:
The old index for the VuePress site has these many records:
![image](https://github.com/grammyjs/website/assets/12952387/8c906dea-0368-43bc-b4a7-a6407763dd94)
The new config and the VitePress site only has these many:
![image](https://github.com/grammyjs/website/assets/12952387/154a9b4a-b5f6-4baf-8e92-7942a761b6e6)
So for some reason it does not find all the content. I am not sure why.

CikiMomogi:
Pretty much the same:

[...]

PR #861, opened 2023-07-05
Use Algolia search

CikiMomogi:
https://github.com/grammyjs/docs-bot/issues/21


CikiMomogi:
Mmm.. Why is it no returning any result?

CikiMomogi:
Maybe due to this https://github.com/grammyjs/website/issues/855?

KnorpelSenf:
No, the index just isn't updated anymore. It's still there. For example, the bot still works.

KnorpelSenf:
Did the search just got a lot faster? This seems to work much better than the internal search (on my device)

CikiMomogi:
That's possible. Maybe the build version is more optimized for Algolia.

CikiMomogi:
If only we could solve the limit result problem (currently max 5 results) and show multi-level results like vuepress does, I would prefer choose Algolia as the main search engine.

KnorpelSenf:
> If only we could solve the result limit problem (currently max 5 results) and have ability to split the results based on each section like vuepress does, I would prefer choose Algolia as the main search engine.

I think it should be possible to add this to vitepress sooner or later.

site/docs/.vitepress/config.ts
```diff
@@ -22,24 +22,17 @@ export default defineConfig({
     siteTitle: "grammY",
     externalLinkIcon: true,
     search: {
-      provider: "local",
+      provider: "algolia",
       options: {
+        apiKey: "33782ffb584887e3b8cdf9e760ea8e60",
+        indexName: "grammy",
+        appId: "RBF5Q0D7QV",
```
CikiMomogi, line 29: Wouldn't it be better to hide these two tokens? It could also educate others to not write credentials in the code directly.

KnorpelSenf: No. Those values are delivered to the client. They are publicly available. It is not possible to hide them.

KnorpelSenf:
What is missing here?

KnorpelSenf (changes requested):
Please add a text file to this repo which contains the Algolia config that we use. Similarly to the IV rules, it will not be synced with Algolia automatically. However, the config is necessary in order to build the site locally and it's somehow part of our logic.

Otherwise, LGTM.

CikiMomogi:
Just to be clear, I should create two files: one with our Algolia crawler config similar to [this one](https://github.com/grammyjs/website/issues/855#issuecomment-1627723698), and another with the IV rules from the [following repository](https://github.com/grammyjs/docs-bot/blob/main/iv-rules.txt), correct?

KnorpelSenf:
No, just one file with the config. The IV rules are in the docs bot repo. You can update them there. I was just giving an example :)

KnorpelSenf (approved):
Thank you so much!

KnorpelSenf:
@roj1512 your turn

roj1512 (approved):
I don’t know if I know anything about this, but I know that I trust you.

KnorpelSenf:
whoop whoop

PR #870, opened 2023-07-08
Never use empty lines around tips

KnorpelSenf:
This was done inconsistently across all languages and within every language.

From now on, we use
- no blank lines inside a tip next to the boundaries
- blank lines surrounding a tip
- blank lines inside a tip surrounding code snippets, even if they are next to a tip
- blank before the closing tag of the tip if the tip content ends with a list entry
- blank before the closing tag of the tip if the tip content ends with a quote

The same applies for other `:::` environments, not just tips.


roj1512 (approved):

KnorpelSenf:
I was faster

KnorpelSenf:
But how did you commit the same changes?

roj1512:
I have null idea

KnorpelSenf:
GH race condition

PR #871, opened 2023-07-10
Update link checker

CikiMomogi:


deno.jsonc
```diff
@@ -1,6 +1,7 @@
 {
   "tasks": {
-    "check-links": "deno run --allow-read --allow-net https://raw.githubusercontent.com/grammyjs/link-checker/main/main.ts site/docs"
+    "clear-cache": "deno cache --reload https://raw.githubusercontent.com/grammyjs/link-checker/anchor/main.ts",
+    "check-links": "deno task clear-cache; deno run --allow-read --allow-net https://raw.githubusercontent.com/grammyjs/link-checker/anchor/main.ts --clean-url --dir=site/docs"
```
roj1512, line 4: I prefer merging these two into one task. I also don’t believe the `--reload` is used that often to have it in a task. What do you think?

CikiMomogi: > I also don’t believe the `--reload` is used that often to have it in a task.

Unfortunately I couldn't find a better way to update the files if there are changes in the link-checker repo. Do you have a better idea?

KnorpelSenf: Provide `-r` as an argument when running the task would allow us to remove `--reload` from the task itself, not sure if that's what we want

CikiMomogi: That works too. Should we continue, @roj1512?

roj1512: Yes.

roj1512 (approved):

PR #877, opened 2023-07-13
Added secret token guide for webhooks

angelhodar:
Hey, this morning I had to transform my long polling bot to use webhook and in the telegram group some people adviced me to use the secret token to verify calls, so I thought it should be not only specified in the API reference :)


KnorpelSenf (changes requested):
We should make clear that this is only necessary if there's no secret path already.

angelhodar:
Shouldn't the secret token option be the recommended one?

KnorpelSenf:
Reopened from a new branch at #880

PR #878, opened 2023-07-15
Even grannY can rock grammY! 🤘

PonomareVlad:
image

### DIFF
--- a/site/docs/README.md
+++ b/site/docs/README.md
@@ -13,6 +13,7 @@ hero:
     - can do anything except dishes.
     - easy peasy lemon squeezY.
     - billions and billions served.
+    - even grannY can rock grammY!
   image:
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
What does this even mean? 🫠

KnorpelSenf:
If the joke is related to https://t.me/grammyjs/143749 then I don't like it lol, I wouldn't want to display a disadvantage of the name that prominently

PonomareVlad:
Hmm, that’s right, but my comedy point was related to the play of words in the context of ease of use and an unexpected image of a granny as a developer 🌚

PonomareVlad:
If there is no value in this PQ, I will not prevent its closure 🙂

roj1512:
Let me take the lead.

PR #880, opened 2023-07-16
Secret token for webhooks

angelhodar:
Sorry didnt know that renaming branch closed PR


KnorpelSenf (changes requested):
> https://github.com/grammyjs/website/pull/877#issuecomment-1637028183

No. From a security point of view, using a token in the header isn't anymore secure than using a token in the header. It's just useful if you want to provide a third-party service access to your bot.

roj1512:
Closing due to inactivity.

PR #781, opened 2023-05-27
Improve VPS hosting guide, add Hostinger to the hosting comparison

niusia-ua:

PR #881, opened 2023-07-26
Update code examples to use new Deno APIs

roj1512:


roziscoding (approved):
LGTM

site/docs/hosting/fly.md
```diff
@@ -21,15 +21,14 @@ You can run your bot using both [webhooks or long polling](../guide/deployment-t
...
-import { serve } from "https://deno.land/std/http/server.ts";
 import { webhookCallback } from "https://deno.land/x/grammy/mod.ts";
 // You might modify this to the correct way to import your `Bot` object.
 import { bot } from "./bot.ts";
 
 const port = 8000;
 const handleUpdate = webhookCallback(bot, "std/http");
 
-serve(async (req) => {
+Deno.serve(async (req) => {
```
dcdunkan, line 31: The server options is the first argument in `Deno.serve` API. So, that `{ port }` (the second argument here) needs to go up.

site/docs/hosting/supabase.md
```diff
@@ -49,7 +48,7 @@ bot.command("ping", (ctx) => ctx.reply(`Pong! ${new Date()}`));
 
 const handleUpdate = webhookCallback(bot, "std/http");
 
-serve(async (req) => {
+Deno.serve(async (req) => {
```
dcdunkan, line 51: This is wrong, right? The server handler must return a `Response`, but it looks like in `catch` it doesn't.

dcdunkan (changes requested):

KnorpelSenf:
I think this is what I meant in #781

PR #909, opened 2023-08-30
[Ukrainian] Improve menu ID

Nazar-Ant:
https://t.me/grammyjs/161795

Ready for merge


Nazar-Ant:
Come on. It came as news to me that the Cyrillic alphabet is encoded in four bytes, not two, like the Latin alphabet. And it's hard to ask such trivial questions when you don't suspect there might be a problem.

KnorpelSenf:
I knew all of these things and even factored this in when implementing the menu. I remember thinking, well, it will now support all of Unicode, but it would be better to stick with ASCII due to the low number of bytes available. I was just being too careless in my review :)

PR #932, opened 2023-10-26
Update getting-started.md

zivc:
Provided NPM instructions do not persist to a package.json.

Added an important step to initialise the npm package.json before installing prescribed modules.


KnorpelSenf:
This is superfluous. If you run `npm i -D typescript`, a `package.json` file will be generated.

PR #947, opened 2023-11-21
Add Algolia crawler config

KnorpelSenf:
Committed here for transparency.


roj1512 (approved):

quadratz:
Actually, we've already uploaded them here:
https://github.com/grammyjs/website/blob/main/site/docs/.vitepress/configs/algolia/crawler.js

IIRC, the main difference is, that one, addresses the issue where the hierarchy results are not appearing.

KnorpelSenf:
I was not aware of this, I will merge the two files in a subsequent PQ and remove this one.

PR #948, opened 2023-11-21
Do not display private props and methods

KnorpelSenf:
We should not list private things in the ref


KnorpelSenf:
It also adds missing docs and link to source for constructors

site/api/components/Class/Constructors.tsx
```diff
@@ -13,13 +15,15 @@ export function Constructors({
...
   }
-  const items = ctors.map(({ accessibility, name, params }, i) => (
+  return ctors.map((v) => (
     <>
-      {accessibility ? `${accessibility} ` : undefined}
-      <span style="color: rgb(98, 232, 132);">{name}</span>(
-      <Params getLink={getLink}>{params}</Params>);
-      {"\n"}
+      <CodeBlock>
+        {v.accessibility ? `${v.accessibility} ` : undefined}
```
roj1512, line 21: v.accessibility should be inside StyleKw

KnorpelSenf: Right

site/api/components/Class/Constructors.tsx
```diff
@@ -13,13 +16,15 @@ export function Constructors({
...
   }
-  const items = ctors.map(({ accessibility, name, params }, i) => (
+  return ctors.map((v) => (
     <>
-      {accessibility ? `${accessibility} ` : undefined}
-      <span style="color: rgb(98, 232, 132);">{name}</span>(
-      <Params getLink={getLink}>{params}</Params>);
-      {"\n"}
+      <CodeBlock>
+        {v.accessibility ? <StyleKw>`${v.accessibility} `</StyleKw> : undefined}
```
roj1512, line 22:
```suggestion
        {v.accessibility ? <StyleKw>{v.accessibility}{" "}</StyleKw> : undefined}
```

KnorpelSenf: That would still style the space with the wrong colour … is there anything else that this improves?

roj1512: @KnorpelSenf No one sees the color of the spaces. You should fix it all other places if you are able to see the color of the spaces.

KnorpelSenf:
@roj1512 please go ahead and merge this if you're okay with it

PR #952, opened 2023-11-22
[Ukrainian] Create demo pages

niusia-ua:
@KnorpelSenf Does "Live Browser Demo" mean that it is something that is open **in** the browser and can be **tapped**? If so, I think that "Interactive Browser Demo" would be a better option, but this is a Ukrainian language option.


roj1512 (approved):
LGTM.

PR #953, opened 2023-11-22
[Ukrainian] Translate untranslated strings

niusia-ua:


site/docs/.vitepress/configs/locales/uk.ts
```diff
@@ -513,7 +513,7 @@ export const notFoundUk: Record<string, NotFound> = {
   uk: {
     title: "PAGE NOT FOUND",
```
quadratz, line 514: This too?

quadratz (changes requested):

site/docs/.vitepress/configs/locales/uk.ts
niusia-ua: I think not, because absolutely everyone knows this phrase. Everyone also knows the 404 number. So in any case, the user understands what happened. Although @deptyped and @LWJerri may have a different opinion?

roj1512:
@niusia-ua Label this PQ.

site/docs/.vitepress/configs/locales/uk.ts
LWJerri: 50/50 in my opinion, but we also must remember - docs can read people who just started learning programming and English, so...

roj1512 (approved):

PR #954, opened 2023-11-23
Cancel pending builds when a commit is pushed

roj1512:


KnorpelSenf (approved):
I'm assuming that this works per branch so it LGTM

PR #959, opened 2023-12-11
Update dependencies and localize the 404 page

quadratz:
Changelog:
1. Updated VitePress to version `1.0.0-rc.31`.
2. Custom-container (tip, warning, danger, code snippet) and button now adheres to the upstream design (ex: no border line).
3. Code snippet theme now utilizes the default theme (no longer using the Dracula theme) as it's more readable in light mode.
4. If a page is not found, it will be redirected to 404.html. This is done as a workaround for pages (especially the navbar menu) that do not adhere to the i18n format.


site/netlify.toml
```diff
@@ -22,3 +22,28 @@ status = 301
...
+
+[[redirects]]
+from = "/uk/*"
+to = "/uk/404.html"
+status = 404
+
+[[redirects]]
+from = "/zh/*"
+to = "/zh/404.html"
+status = 404
```
quadratz, line 49: Are these configs correct? I got them from https://docs.netlify.com/routing/redirects/redirect-options/?#custom-404-page-handling. But it seems the 404 page is still not being redirected. What's wrong? :thinking:

quadratz:
Ready for review

KnorpelSenf:
> 4\. If a page is not found, it will be redirected to 404.html. This is done as a workaround for pages (especially the navbar menu) that do not adhere to the i18n format.

Can you give an example for when this happens?

site/docs/.vitepress/deps/langs.ts
```diff
@@ -0,0 +1,38 @@
...
+          page.value.relativePath.slice(currentLang.value.link.length - 1),
+          !site.value.cleanUrls,
+        ),
+      }
+    )
+  );
+  return { localeLinks, currentLang };
+}
+// @ts-ignore
+function normalizeLink(link, addPath, path, addExt) {
```
KnorpelSenf, line 29: Please add a comment explaining what normalisation is done and why

KnorpelSenf:
Feel free to update the title once more if you feel like its inaccurate

quadratz:
> > 4. If a page is not found, it will be redirected to 404.html. This is done as a workaround for pages (especially the navbar menu) that do not adhere to the i18n format.
> 
> Can you give an example for when this happens?

https://grammy.dev/id/foo, the navbar at the top remains in English instead of being in Indonesian as intended.

site/docs/.vitepress/deps/langs.ts
quadratz: The files in the 'deps' folder are from Vitepress. Since I couldn't import them directly, I duplicated the files into that directory and imported them from there.

KnorpelSenf: Wait so you didn't write the code in the deps folder? That sounds a lot like we should rather find a way to import the functionality, not copy the build output into this repo.

KnorpelSenf:
Oooh I see, I didn't even notice! That makes a lot more sense now, thanks

site/docs/.vitepress/deps/langs.ts
KnorpelSenf: Maybe I need to take a second look to understand more about this

quadratz: Yes, please. :')

KnorpelSenf (approved):
LGTM, thanks!

roj1512 (approved):
LGTM.

PR #977, opened 2024-01-02
Fix typos

quadratz:


roj1512 (approved):
I would have forgotten this, thank you so much!

PR #980, opened 2024-01-04
Improve headings

roziscoding:
Previous titles were a bit weird and misleading since the examples don't really show any "routes"


roj1512 (commented):
Is this really not translatable?

roziscoding:
To be fair, I assumed it isn't, but I don't have the knowledge of other languages to check their versions so I guess it's better to remove the label

roj1512 (approved):

niusia-ua:
@LWJerri @deptyped Do we need to correct this? In our translation, we used the word "обробник" (in English, "handler"), which is essentially the same as "route". Based on the meaning of the text and code examples, I don't think we need to change anything.

LWJerri:
> обробник

In my opinion, we can do translating to "В разовому випадку" та "У всіх випадках". But yes, my suggestion is not word-to-word translation with original file, so idk.

LWJerri (approved):

roj1512 (approved):

habemuscode (approved):

PR #989, opened 2024-02-11
Update deps

KnorpelSenf:
still not signed
no idea why


KnorpelSenf:
@roj1512 GPG signing seems to work, don't ask me why. I'll stick with that now.

roj1512 (approved):

PR #996, opened 2024-02-18
Update deps

quadratz:
- Update dependencies.
- Remove memory allocation flag (https://github.com/grammyjs/website/commit/d73e787586925ac0a8ba49a924ccd005c9c23ae8).
  This flag was inherited from Vuepress. Seems we no longer face memory constraints in Vitepress. The build runs smoothly without requiring an expanded memory allocation size.
- Use exact versions for all dependencies.


KnorpelSenf (approved):
Thank you so much!

PR #1005, opened 2024-03-07
Fix typos

KnorpelSenf:
wrong comma, "doing any X not Xs"


roj1512 (approved):
LGTM. Why is "any actions" wrong? Just a personal doubt.

KnorpelSenf:
> Why is "any actions" wrong?

It just felt right.

My intuition is that “any actions” means “we all know which actions we are talking about, and zero of them should happen before the check.” In contrast, “any action” means “whatever type of action you would like to perform, don't do it before the check.” The singular form therefore emphasizes that the type of action is unknown, while the plural form would rather refer to the number of actions.

Both versions work, though, and I guess you will also find people who argue differently.

PR #1010, opened 2024-03-17
Generate API reference on npm install

KnorpelSenf:
Just like we index versions upon install, we should generate the API once


roj1512 (approved):

roj1512 (approved):

KnorpelSenf:
classic merge before CI, love it

PR #1023, opened 2024-03-25
Add ranking system to Algolia search

roziscoding:


KnorpelSenf:
That feels like it should work. @quadratz can you test this again review the PQ?

quadratz:
Not good. They throw an error if there is any line of code other than `new Crawler(...)`, and the `pageRank` property is not available.

```log
[Error] parsing_error
Crawler is not defined
raw object 
{success: false, code: 400, message: "parsing_error", details: Array(1), lineNumber: null}
success: false
code: 400
message: "parsing_error"
details: (1) [Object]
0: Object
label: "config"
type: "format"
message: "Crawler is not defined"
```

quadratz:
Maybe we can inject `searchRank="-1"` attributes into every page of `ref`. I will look into them later.

KnorpelSenf:
> Maybe we can inject `searchRank="-1"` attributes into every page of `ref`. I will look into them later.

Is this something you'd want to include in #1024?

quadratz:
Keeping the work in this PQ is better, I think.

quadratz:
Review please --> https://cm-grammy.netlify.app/

roziscoding:
Awesome!

KnorpelSenf:
@quadratz IT HAS SECTIONS? DAMN that's so sweet, let's get that juicy junk merged right awaY!

site/docs/.vitepress/configs/algolia/crawler.js
```diff
@@ -18,7 +18,7 @@ const getPageRank = (url: Location) => {
 
 new Crawler({
   appId: "RBF5Q0D7QV",
-  apiKey: "...",
+  apiKey: "33782ffb584887e3b8cdf9e760ea8e60",
```
KnorpelSenf, line 21: Please revert

quadratz: Why?

KnorpelSenf: It's the key from your installation, not the one grammY uses. It shouldn't be part of this repo.

quadratz: It's the same API we use in https://github.com/grammyjs/website/blob/main/site/docs/.vitepress/configs/algolia/index.ts

KnorpelSenf: Oh, then it's the frontend key. Either way, it cannot be used for the crawler since it doesn't have write access to the index.

quadratz: Okay

KnorpelSenf (approved):
Lessgo!

site/docs/.vitepress/configs/algolia/crawler.js
quadratz: In my case I have the same API key for both frontend and crawler.

quadratz: But not the write access / admin API key

KnorpelSenf: Huh, I don't 🫠

KnorpelSenf: No idea why. They once changed a lot of things and now I no longer dare to touch anything as long as things are working.

KnorpelSenf:
@roj1512 we're ready to have this merged now 😅

PR #1028, opened 2024-03-27
Translate `docFooter`

niusia-ua:
These interface elements are translated into Indonesian only. I think other translators need to decide whether to translate this.


niusia-ua:
image_2024-03-27_05-23-27

PR #1030, opened 2024-03-28
Update dependencies

quadratz:
Changes:
- Update dependencies.
- Remove package-lock.json to avoid conflicts with Windows setup due to different dependencies.
As we already use pinned versions in `package.json`, this should not cause any issues.


KnorpelSenf:
> * Remove package-lock.json to avoid conflicts with Windows setup due to different dependencies.
>   As we already use pinned versions in `package.json`, this should not cause any issues.

Can you elaborate on this issue? Shouldn't npm work identically on Windows, too?

Running `npm i` instead of `npm ci` is quite a bit slower. This means that CI will take more time, so we have to wait even longer for every deployment preview to show up. If there's a way to keep the lock file, I'd like to keep it.

roj1512:
I personally don’t use Windows, I don’t think @KnorpelSenf does either. Why don’t we instead have a note for Windows users?

KnorpelSenf:
A fix for windows > a note for windows > no package lock and worse perf for everybody

quadratz:
> Can you elaborate on this issue? Shouldn't npm work identically on Windows, too?
```log
PS C:\Users\qz\dev\grammy\website\site> npm run docs:build

> docs:build
> npm run docs:vitepress && npm run docs:sitemap


> docs:vitepress
> vitepress build docs

C:\Users\qz\dev\grammy\website\site\node_modules\rollup\dist\native.js:87
                throw new Error(
                      ^

Error: Cannot find module @rollup/rollup-win32-x64-msvc. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at requireWithFriendlyError (C:\Users\qz\dev\grammy\website\site\node_modules\rollup\dist\native.js:87:9)  
    at Object.<anonymous> (C:\Users\qz\dev\grammy\website\site\node_modules\rollup\dist\native.js:96:76)       
    ... 3 lines matching cause stack trace ...
    at Module._load (node:internal/modules/cjs/loader:1028:12)
    at cjsLoader (node:internal/modules/esm/translators:359:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:308:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24) {
  [cause]: Error: Cannot find module '@rollup/rollup-win32-x64-msvc'
  Require stack:
  - C:\Users\qz\dev\grammy\website\site\node_modules\rollup\dist\native.js
      at Module._resolveFilename (node:internal/modules/cjs/loader:1149:15)
      at Module._load (node:internal/modules/cjs/loader:990:27)
      at Module.require (node:internal/modules/cjs/loader:1237:19)
      at require (node:internal/modules/helpers:176:18)
      at requireWithFriendlyError (C:\Users\qz\dev\grammy\website\site\node_modules\rollup\dist\native.js:69:10)
      at Object.<anonymous> (C:\Users\qz\dev\grammy\website\site\node_modules\rollup\dist\native.js:96:76)     
      at Module._compile (node:internal/modules/cjs/loader:1378:14)
      at Module._extensions..js (node:internal/modules/cjs/loader:1437:10)
      at Module.load (node:internal/modules/cjs/loader:1212:32)
      at Module._load (node:internal/modules/cjs/loader:1028:12) {
    code: 'MODULE_NOT_FOUND',
    requireStack: [
      'C:\\Users\\qz\\dev\\grammy\\website\\site\\node_modules\\rollup\\dist\\native.js'
    ]
  }
}

Node.js v21.6.1
```

> Running npm i instead of npm ci is quite a bit slower. This means that CI will take more time, so we have to wait even longer for every deployment preview to show up. If there's a way to keep the lock file, I'd like to keep it.

TIL package-lock.json may improve performance. If its removal results in a significant performance drop, we should add notes for windows contributors to use `npm i --no-package-lock` instead.

> I personally don’t use Windows, I don’t think @KnorpelSenf does either. ...

I understand your point. However, that way of thinking may discourage new contributors who are not using the same operating system as you.

roj1512:
@quadratz Please see #1031.

quadratz:
Can you push the package-lock.json? Mine is Windows so it is not compatible.

roj1512:
Why don’t we even use pnpm

quadratz:
Yes, please.

KnorpelSenf:
I still have this dream that we can just use Deno and forget about node_modules entirely but last time I checked, it didn't manage to run vitepress

roj1512 (approved):
Thank you.

PR #1033, opened 2024-04-01
Minor fixes after the 7.2 compat release

KnorpelSenf:


site/docs/guide/filter-queries.md
```diff
@@ -9,7 +9,7 @@ Other filtering of messages is left to the developer, which often leads to endle
 
 On the contrary, **grammY ships with its own query language** that you can use in order to **filter for exactly the messages** you want.
 
-This allows for over 820 different filters to be used, and we may add more over time.
+This allows for over 1150 different filters to be used, and we may add more over time.
```
roj1512, line 12: Is 1150 an estimate or the exact count?

KnorpelSenf: The exact count is 1153, but I don't want to give the exact number in case we have to do minor changes to the implementation. I don't want to maintain it that closely.

roj1512: LGTM then. Writing 1100 would look better imo, but it is not critical.

roj1512 (approved):

LWJerri (approved):

quadratz (approved):

site/docs/id/guide/context.md
```diff
@@ -55,15 +54,17 @@ Object context selalu berisi informasi tentang bot-mu, yang dapat diakses melalu
...
-| `ctx.entities`        | Mendapatkan entity pesan beserta teksnya, dapat disaring berdasarkan jenis entity-nya      |
-| `ctx.reactions`       | Mendapatkan reaksi dari suatu update [dengan mudah](./reactions#menyimak-perubahan-reaksi) |
+| Shortcut                   | Deskripsi                                                                                  |
+| -------------------------- | ------------------------------------------------------------------------------------------ |
+| `ctx.msg`                  | Mendapatkan object message, termasuk yang sudah diedit                                     |
+| `ctx.chat`                 | Mendapatkan object chat                                                                    |
+| `ctx.senderChat`           | Mendapatkan object chat pengirim dari `ctx.msg` (untuk pesan grup/channel anonim)          |
+| `ctx.from`                 | Mendapatkan informasi penulis pesan, callback query, dan lainnya                           |
+| `ctx.msgId`                | Mendapatkan id pesan untuk memproses pesan atau reaksi                                     |
```
dzikrl, line 63: > Mendapatkan id pesan untuk **memproses** pesan atau reaksi

What does `memproses` do here?
So, `reaksi` need to add `memberi reaksi`?

site/docs/id/guide/context.md
```diff
@@ -55,15 +54,17 @@ Object context selalu berisi informasi tentang bot-mu, yang dapat diakses melalu
...
+| `ctx.inlineMessageId`      | Mendapatkan id pesan inline dari callback query atau hasil inline yang dipilih             |
+| `ctx.businessConnectionId` | Mendapatkan id koneksi bisnis untuk memproses pesan atau update koneksi bisnis             |
```
dzikrl, line 65: > Mendapatkan id koneksi bisnis untuk **memproses** pesan atau update koneksi bisnis

And this too.

site/docs/id/guide/context.md
KnorpelSenf: > But, `ctx.msgId` isn't a method right?

It is a getter, i.e. a property that will run some code when accessed. You do not have to do `ctx.msgId()`, and it is best to think about it as a regular property like `ctx.update`.

dzikrl: Oh, I see...

KnorpelSenf: (in fact, all of the things listed in this table work that way, with the notable exceptions being `ctx.reactions()` and `ctx.entities()`

dzikrl: I took a quick look at the code in `context.ts`, and sure enough it's a getter.

I never use getter or setter when creating classes. Thanks, I now understand when it needs to be used.

dzikrl (approved):
OK.

quadratz (approved):

PR #1042, opened 2024-04-10
Use `performance.now()` instead of `Date.now()`.

LWJerri:


KnorpelSenf:
This is educational code. Its purpose is to be understood by as many people as possible, not to be useful in production. `performance.now` is a much lesser known API than `Date.now`, so these changes make the docs worse. 

The changes would probably be very useful if you had a bot that required to measure time with sub-millisecond accuracy. That's not the goal here, though.

PR #1051, opened 2024-05-15
Translate the entire website interface

niusia-ua:
In fact, almost everything has already been translated. I have corrected some things in the Ukrainian and Spanish versions (based on the translation provided). The Indonesian version of the site has already been fully translated. There are only a few possible changes left in the Chinese version, which I have marked with the comment `//!`. If some of this is not worth translating, I think it is better to leave a comment so that we do not return to it again.

Translated with DeepL.com (free version)

### DIFF
--- a/site/docs/.vitepress/configs/locales/es.ts
+++ b/site/docs/.vitepress/configs/locales/es.ts
@@ -447,7 +447,7 @@ export const siteEs: LocaleConfig = {
       },
       outline: {
         level: [2, 6],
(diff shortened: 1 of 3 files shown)


KnorpelSenf (commented):
Very cool changes

KnorpelSenf (approved):

roj1512 (approved):

KnorpelSenf:
@roj1512 What's blocking the merge?

PR #1053, opened 2024-05-15
Update i18n.md with information about Unicode isolation marks

alyavasilyeva:
Provide information about possible formatting issues when using placeables and how to solve them.

Context: https://github.com/grammyjs/i18n/issues/47

PR #1058, opened 2024-05-21
Update i18n docs with information about Unicode isolation marks (#1053)

KnorpelSenf:
Supersedes #1053.


dcdunkan (commented):
This link only seem to have been added to the Indonesian translation.

https://github.com/projectfluent/fluent.js/wiki/Unicode-Isolation

quadratz:
Yes, that's intentional (https://github.com/grammyjs/website/pull/1053#discussion_r1604214506). Should I remove them?

dcdunkan:
> Yes, that's intentional (https://github.com/grammyjs/website/pull/1053#discussion_r1604214506). Should I remove them?

I think we should keep it as it'll be useful for the readers.

PR #1062, opened 2024-05-23
Fix country flag emoji not displayed correctly on chromium based browsers

quadratz:
This PR fixes the issue of the country flag emoji not being displayed correctly on chromium based browsers. The issue is caused by a problem with the chromium itself [https://issues.chromium.org/issues/40766658](https://issues.chromium.org/issues/40766658).

Example:
https://grammy.dev/hosting/comparison#vps

Before:
screenshot_before

After:
Screenshot_after


roj1512 (approved):
Thank you so much!

PR #1064, opened 2024-05-25
Recommend setting WorkingDirectory

niusia-ua:


site/docs/hosting/vps.md
```diff
@@ -201,6 +201,49 @@ systemd is a powerful service manager which is pre-installed on many Linux distr
    - `WantedBy=multi-user.target` --- defines the system state in which the service should be launched.
      `multi-user.target` --- is a typical value for servers.
 
+   :::: tip Using the `.env` file
```
KnorpelSenf, line 204: I think we're using title case for tips, so it should be
```suggestion
   :::: tip Using the `.env` File
```

niusia-ua: Yes, yes. I always forget about it 😅

roj1512 (commented):
I have nothing against these changes. But why don’t we just suggest specifying WorkingDirectory by default? Who wants the working directory of their bot to be like `/root`, etc.?

niusia-ua:
Oh, that's brilliant. I hadn't even thought of that.
In this case, I'll specify the WorkingDirectory and give two ways to load environment variables.

roj1512 (changes requested):
Looks good. But the section about the start command should be edited since it is no longer required for the script's path to be absolute.

For example, it can be just

```
/home/user/.deno/bin/deno run --allow-all mod.ts
```

NB: You are also missing a `run` in the Deno example.

roj1512 (approved):

site/docs/hosting/vps.md
```diff
@@ -143,18 +143,20 @@ systemd is a powerful service manager which is pre-installed on many Linux distr
...
+
+   # Path to the bot directory: /home/user/bot1/
 
    # Deno example:
-   # /home/user/.deno/bin/deno --allow-all /home/user/bot1/mod.ts
+   # /home/user/.deno/bin/deno --allow-all run ./mod.ts
 
    # Node.js example:
-   # /home/user/.nvm/versions/node/v16.9.1/bin/node /home/user/bot1/index.js
+   # /home/user/.nvm/versions/node/v16.9.1/bin/node ./index.js
```
roj1512, line 159:
```suggestion
   # /home/user/.nvm/versions/node/v16.9.1/bin/node index.js
```

roj1512 (approved):

LWJerri (approved):
:shipit:

PR #1067, opened 2024-05-29
Fix syntax highlighting light theme in API ref

quadratz:
The syntax color in the ref docs (dracula-theme) is hard to read in light mode and it does not match the current theme (github-theme). This PQ should resolve the issue.

Before (dracula-theme):
before
After (github-theme):
after

### DIFF
--- a/site/api/components/Class/Constructors.tsx
+++ b/site/api/components/Class/Constructors.tsx
@@ -24,7 +24,9 @@ export function Constructors({
             {v.accessibility
               ? <StyleKw>{v.accessibility}{" "}</StyleKw>
               : undefined}
-            <span style="color: rgb(98, 232, 132);">{v.name}</span>(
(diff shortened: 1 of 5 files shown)


KnorpelSenf:
> it does not match the current theme (github-theme)

And by current theme you mean the current theme for code snippets in light mode for the rest of the website?

roj1512 (approved):
Thank you so much! I had this in mind but never knew what the best practice would be.

quadratz:
> And by current theme, you mean the current theme for code snippets in light mode for the rest of the website?

Yes. Previously, it only used one theme, which was designed for dark mode only. Now it should use two different themes for both light mode and the dark mode, like we already do with the rest of the website.

quadratz:
> Thank you so much! I had this in mind but never knew what the best practice would be.

I also feel uncomfortable with hardcoding the values. It should be done dynamically. But at least this is what we can do for now.

PR #1068, opened 2024-05-29
Update menu.md with info on allowed_updates

garmeeh:
Couldn't see this anywhere in the docs and wasn't immediately obvious to someone starting to use grammy and Telegram for the first time. 

Feel free to close if it actually just confuses things.


KnorpelSenf:
I see the point of adding such a note, but I'd like to do it slightly differently. Does #1069 work for you?

garmeeh:
Thats perfect @KnorpelSenf 👌

PR #1069, opened 2024-05-29
Add note about `allowed_updates` to the menu plugin page

KnorpelSenf:
Supersedes and thus closes #1068.

- allowed updates can be specified not only for `bot.start()` but also for webhooks and with grammY runner
- this should only be an optional note (not a text paragraph), most people don't configure this
- we already have a note about callback query updates in the middleware system, so it makes sense to put these two hints close to each other


KnorpelSenf:
Approval in https://github.com/grammyjs/website/pull/1068#issuecomment-2137372754

roj1512 (approved):

PR #1087, opened 2024-07-05
[Translate] [Russian] Translate website into Russian

MasedMSD:
In progress.
I will be very happy if you accept this translation when it is completed
My translation may not be perfect, so I would welcome corrections


KnorpelSenf:
@MasedMSD are you no longer interested in this? That was a short endeavor :D

MasedMSD:
@KnorpelSenf 
still interested. I wanted to organize by commits, but then I saw that for this you need to create a issue for that, so I've decided to finish the translation for now, then do it by the book.

MasedMSD:
@KnorpelSenf Is it mandatory to create a translation with issue? I was thinking of creating a branch and putting all translations there so that other users can help as well

KnorpelSenf:
Nice! Are you aware that there has been a previous attempt at translating the docs to Russian? The first attempt failed and the work was abandoned, but perhaps you still want to check it out and see if bringing the old docs up to date is easier than doing everything from scratch.

MasedMSD:
Great idea, but I can't find the source of this translation

PR #1094, opened 2024-07-16
Fix link issues

KnorpelSenf:


KnorpelSenf:
The link checker issue was fixed. Since we never bothered to pin the version of that tool, it should just work now. @rojvv perhaps we could now also make it a required check for all PQs?

I'll look into making link checking faster by fixing strom in the future. That should bring down CI times by a lot.

KnorpelSenf:
@rojvv OMG CI PASSED WITH ALL LINKS CHECKED MERGE NOW BEFORE IT IS TOO LATE THIS IS THE FIRST TIME IN YEARS WE WILL STILL OUR GRANDCHILDREN ABOUT THIS DAY NOW DO NOT LOSE TIME

rojvv (approved):

PR #1098, opened 2024-07-19
Fix #1097

### LINKED ISSUE #1097: Wrong keyboard type in plugins/inline-query
Most likely the link should ref to `./keyboard#building-an-inline-keyboard`

https://github.com/grammyjs/website/blob/2dd2a02228cd2fafd9760ab6b29cc2acafda5fa3/site/docs/plugins/inline-query.md?plain=1#L215

MasedMSD:
Just a little fix to ref

Closes #1097


KnorpelSenf:
Awesome, thanks! Do you also want fix up the other two languages?

MasedMSD:
Yeah, looks like espanol have this misspell tho

MasedMSD:
Oops, alr, looks like Ukrainian was right

KnorpelSenf (approved):
Now I see why only 3/5 langs are fixed, Chinese and Indonesian never had the problem. Thanks!

rojvv (approved):

PR #1102, opened 2024-07-21
Add alias to deployments

rojvv:


KnorpelSenf:
This removes the unique deployment URLs. They have been invaluable in the past, so I am not ready to give them up just yet.

KnorpelSenf:
According to my research, Netlify simply does not support this. I have opened https://github.com/nwtgck/actions-netlify/issues/1187 but I do not expect much from it. We may have to upload all data twice, or maybe switch to a different hosting provider that supports this.

PR #1105, opened 2024-07-30
feat: add bun related docs

stonega:

PR #1129, opened 2024-10-01
feat: bun related docs (@stonega)

MasedMSD:
Originally [PR](https://github.com/grammyjs/website/pull/1105) was created by @stonega, but it looks like he's not going to continue with this PR. So, I think we should mention Bun as a possible way to work with a bot, even though Bun has a large number of issues. Also Bun finally supports [conversations plugin](https://grammy.dev/plugins/conversations) and a decent part of Russian language chat users use it for their projects.

I made some changes and added Bun mentions to the hosting files documentation

### DIFF
--- a/site/docs/guide/getting-started.md
+++ b/site/docs/guide/getting-started.md
@@ -150,6 +150,90 @@ in your terminal before you execute `node bot.js`.
 This makes it easier to debug your bot.
 :::
 
+## Getting Started on Bun
+
+> This guide assumes that you have [Bun](https://bun.sh) installed.
+> If you don't know what these things are, check out our [Introduction](./introduction)!
+
+Create a new project and install the `grammy` package.
+Do this by opening a terminal and typing:
+
+```sh
+# Create a new directory and change into it.
+mkdir my-bot
+cd my-bot
+
+# Run bun init to scaffold a new project.
+bun init --yes
+
+# Install grammY.
+bun add grammy
+```
+
+Your folder structure should looks like this:
+
+```asciiart:no-line-numbers
+.
+├── index.ts
+├── .gitignore
+├── node_modules/
+├── package.json
+├── bun.lockb
+├── README.md
+└── tsconfig.json
+```
+
+Now, it's time to open Telegram to create a bot account, and obtain a bot token for it.
+Talk to [@BotFather](https://t.me/BotFather) to do this.
+The bot token looks like `123456:aBcDeF_gHiJkLmNoP-q`.
+It is used to authenticate your bot.
+
+Got the token? You can now code your bot in the `index.ts` file.
+You can copy the following example bot into that file, and pass your token to the `Bot` constructor:
+
+```ts [TypeScript]
+import { Bot } from "grammy";
+
+// Create an instance of the `Bot` class and pass your bot token to it.
+const bot = new Bot(""); // <-- put your bot token between the ""
+
+// You can now register listeners on your bot object `bot`.
+// grammY will call the listeners when users send messages to your bot.
+
+// Handle the /start command.
+bot.command("start", (ctx) => ctx.reply("Welcome! Up and running."));
+// Handle other messages.
+bot.on("message", (ctx) => ctx.reply("Got another message!"));
+
+// Now that you specified how to handle messages, you can start your bot.
+// This will connect to the Telegram servers and wait for messages.
+
+// Start the bot.
+bot.start();
+```
+
+You can now run the bot by executing
+
+```sh
+bun run bot.ts
+```
+
+in your terminal.
+Done! :tada:
+
+Head over to Telegram to watch your bot respond to messages!
+
+::: tip Enabling Logging
+You can enable basic logging by running
+
+```sh
+export DEBUG="grammy*"
+```
+
+in your terminal before you execute `bun run index.ts`.
+This makes it easier to debug your bot.
+:::
+
 ## Getting Started on Deno
 
 > This guide assumes that you have [Deno](https://deno.com) installed.
--- a/site/docs/guide/introduction.md
+++ b/site/docs/guide/introduction.md
@@ -66,7 +66,7 @@ You will get to know them as you go.
 
 ## Prerequisites to Getting Started
 
-> Skip the rest of this page if you already know how to develop a Deno or a Node.js application, and [get started](./getting-started).
+> Skip the rest of this page if you already know how to develop Deno, Bun or Node.js application, and [get started](./getting-started).
 
 Here are a few interesting things about programming---things that are essential to coding, yet rarely explained because most developers think they are self-evident.
 
@@ -172,6 +172,34 @@ You can stop the bot again with `Ctrl+C`.
 
 Ready?
 [Get started](./getting-started#getting-started-on-deno)! :robot:
+
+### Prerequisites for Bun
+
+Bun is a new JavaScript runtime, which is another good choice to build bot, like Deno, you are going to write your bot in TypeScript.
+The exact commands for all of that will be introduced in the next section when you actually create a bot, but it is important to know that these steps are necessary.
+
+Firstly, you have to have [Bun](https://bun.sh/) installed.
+
+In summary, this is what you have to do for Bun:
+
+Create a new directory somewhere.
+It will contain your bot project.
+Open this new directory in VS Code.
+
+```sh
+mkdir ./my-bot
+cd ./my-bot
+code .
+```
+
+Then:
+
+1. Run `bun init --yes` in your terminal to initialize the project.
+2. Create a source file `bot.ts` with TypeScript code inside the project.
+3. Run `bun run bot.ts` from your terminal, or run `bun --watch run bot.ts` if you want to keep updated with file changes.
+
+Ready?
+[Get started](./getting-started#getting-started-on-bun)! :robot:
 
 ### Prerequisites for Node.js
 
--- a/site/docs/hosting/firebase.md
+++ b/site/docs/hosting/firebase.md
@@ -49,7 +49,7 @@ npm install -g firebase-tools
    - TypeScript
 6. Optionally, you can select ESLint.
 7. The CLI asks you if you want to install the dependencies with npm.
-   If you use another package manager like `yarn` or `pnpm` you can decline.
+   If you use another package manager like `yarn`, `bun` or `pnpm` you can decline.
    In that case, you have to `cd` into the `functions` directory and install the dependencies manually.
 8. Open `./functions/package.json` and look for the key: `"engines": {"node": "16"}`.
    The `node` version should match your installed version of Node.js.
--- a/site/docs/hosting/vps.md
+++ b/site/docs/hosting/vps.md
@@ -260,6 +260,10 @@ yarn global add pm2
 pnpm add -g pm2
 ```
 
+```sh [bun]
+bun add -g pm2
+```
+
 :::
 
 #### Creating an Application


site/docs/guide/getting-started.md
```diff
@@ -150,6 +150,90 @@ in your terminal before you execute `node bot.js`.
...
+// This will connect to the Telegram servers and wait for messages.
+
+// Start the bot.
+bot.start();
+```
+
+You can now run the bot by executing
+
+```sh
+bun run bot.ts
```
MasedMSD, line 218: Didn't changed that
```
bun run index.ts
```

site/docs/guide/introduction.md
```diff
@@ -173,6 +173,34 @@ You can stop the bot again with `Ctrl+C`.
...
+```sh
+mkdir ./my-bot
+cd ./my-bot
+code .
+```
+
+Then:
+
+1. Run `bun init --yes` in your terminal to initialize the project.
+2. Create a source file `bot.ts` with TypeScript code inside the project.
```
MasedMSD, line 198: Should change `bot.ts` to `index.ts` which bun created by default when using `bun init --yes`

site/docs/guide/getting-started.md
```diff
@@ -150,6 +150,90 @@ in your terminal before you execute `node bot.js`.
...
+Create a new project and install the `grammy` package.
+Do this by opening a terminal and typing:
+
+```sh
+# Create a new directory and change into it.
+mkdir my-bot
+cd my-bot
+
+# Run bun init to scaffold a new project.
+bun init --yes
```
KnorpelSenf, line 167: Does Bun really require a setup script?

site/docs/guide/getting-started.md
KnorpelSenf: Why not?

KnorpelSenf: Actually, it would be much better to create a `bot.ts` file instead of `index.ts`. It makes no sense to do this inconsistently for Bun.

site/docs/guide/introduction.md
KnorpelSenf: How can this be fixed?

site/docs/guide/getting-started.md
MasedMSD: > Why not?

Just forgot about it. These are comments addressed to me in the morning 😶‍🌫️

MasedMSD: > Actually, it would be much better to create a `bot.ts` file instead of `index.ts`. It makes no sense to do this inconsistently for Bun.

Hm, you're right. If all the examples use `bot.ts`, it would be convenient to leave this for association

site/docs/guide/introduction.md
MasedMSD: > How can this be fixed?

Just change filename

site/docs/guide/getting-started.md
MasedMSD: Not really, Bun just uses it for creating default files like README, package.json, tsconfig and index.ts.

I don't know about older versions, but the new ones add the package without init

MasedMSD:
Should we leave `index.ts` or should we tell the user to create a `bot.ts` file each time in order for the NodeJS and Deno examples to be the same?

site/docs/guide/getting-started.md
KnorpelSenf: Do we have to use the init script then? Maybe we can just create the files manually or use `npm init`?

site/docs/guide/introduction.md
KnorpelSenf: Wait so there's no way to skip creating the file, or so create it with a different name?

site/docs/guide/getting-started.md
MasedMSD: Yeah, why not. If the Bun has its own initialization that looks more convenient than npm, why we shouldn't use it

site/docs/guide/introduction.md
MasedMSD: I don't guess. Bun by default create index.ts with content:

```ts
console.log("Hello via Bun!");
```

site/docs/guide/getting-started.md
KnorpelSenf: tsconfig can't be created by bun, unless it installs and runs `tsc`? Because the rest of the website depends on having the TypeScript default config and I don't think I want to trust bun on making sure that these two files are always in sync, so we have to tell people to run `tsc` themselves.

MasedMSD: > tsconfig can't be created by bun, unless it installs and runs `tsc`? Because the rest of the website depends on having the TypeScript default config and I don't think I want to trust bun on making sure that these two files are always in sync, so we have to tell people to run `tsc` themselves.

I don't quite understand the question. Bun does not use tsc for compilation. It runs Typescript immediately

site/docs/guide/getting-started.md
KnorpelSenf: Because it clearly isn't more convenient, it creates the wrong files

MasedMSD: What you mean by `creates the wrong files`

site/docs/guide/introduction.md
KnorpelSenf: I don't see how we need that. Note that we also don't run `deno init` for the same reason.

MasedMSD: > I don't see how we need that. Note that we also don't run `deno init` for the same reason.

This is to show that we are starting a clean project. Of course, we may not init the project and start a ready-made one.

site/docs/guide/getting-started.md
KnorpelSenf: This is incorrect. It should be this:
```json
{
  "compilerOptions": {
    "target": "es2016",
    "module": "commonjs",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "skipLibCheck": true
  }
}
```

MasedMSD: Why?

site/docs/guide/getting-started.md
KnorpelSenf: The files it creates have the wrong content.
- `index.ts` should actually be named `bot.ts`
- `bot.ts` should actually different content
- `tsconfig.json` is wrong
- we have not even talked about package.json, will this also be a problem?

site/docs/guide/introduction.md
KnorpelSenf: Since almost everything `bun init` does is wrong for our use case, I don't understand why it would be useful. The docs would be:
1. Run `bun init`
2. Delete all files it created
3. Create correct files

site/docs/guide/getting-started.md
MasedMSD: - `index.ts` should actually be named `bot.ts` - not a problem.
- `bot.ts` should actually have different content - not a problem again. Of course user should write default code for bot launching
- `tsconfig.json` is wrong - what's wrong with it?
- we have not even talked about package.json, will this also be a problem? - Not actually. package.json is:

```JSON
{
  "name": "test",
  "module": "index.ts",
  "type": "module",
  "dependencies": {
    "grammy": "^1.30.0"
  },
  "devDependencies": {
    "@types/bun": "latest"
  },
  "peerDependencies": {
    "typescript": "^5.0.0"
  }
}
```

MasedMSD: image

site/docs/guide/introduction.md
MasedMSD: It's not deleting already created files

site/docs/guide/getting-started.md
KnorpelSenf: Because the rest of the docs require it. Every single of the ~700 code examples are tested on both Deno and Node and they work with the config created in the getting started guide. So we either make sure that Bun is compatible with the Node examples, or else we will have to re-create all code examples with a Bun tab. Even if it turns out to the mostly the same in most cases, we have to least test everything.

I don't care about bun and I don't want to spend time on this. I am not going install it. I also want to be able to edit the docs myself, and write new docs. This is not possible if I have to test everything on bun first. I won't accept any contributions that prevent me from changing the docs in the future, or that force me to ask for help for every single edit.

The bottom line is that the bun setup MUST work with the existing Node code. It must type-check with the same TS config, and it must run correctly.

MasedMSD: > The bottom line is that the bun setup MUST work with the existing Node code. It must type-check with the same TS config, and it must run correctly.

I understand your indignation about this runtime, but I have not had any problems with it before when working with it.

KnorpelSenf: Actually, nobody will test if it runs correctly, people are just going to watch it fail when they use Bun. But at least we can make sure that it type-checks.

site/docs/guide/getting-started.md
KnorpelSenf: Why is TS a peer dep? It should be a dev dep.

site/docs/guide/introduction.md
KnorpelSenf: Read again

site/docs/guide/getting-started.md
MasedMSD: > This is incorrect. It should be this:
> 
> ```json
> {
>   "compilerOptions": {
>     "target": "es2016",
>     "module": "commonjs",
>     "esModuleInterop": true,
>     "forceConsistentCasingInFileNames": true,
>     "strict": true,
>     "skipLibCheck": true
>   }
> }
> ```

I'll check all examples with Bun if you want so
image

KnorpelSenf: > > The bottom line is that the bun setup MUST work with the existing Node code. It must type-check with the same TS config, and it must run correctly.
> 
> I understand your indignation about this runtime, but I have not had any problems with it before when working with it.

Sure but you also did not test everything

KnorpelSenf:
I have now wasted more time on this than I ever wanted, maybe I'll get back to this in a week or so

site/docs/guide/getting-started.md
MasedMSD: Bun adds a tiзуscript to Peer Deps, but honestly I haven't tried to figure out the behaviour of that. We could move it to dev deps and nothing would change.

KnorpelSenf (changes requested):
The bottom line of this is still that whatever we do with the tutorial, the output must look exactly like https://github.com/grammyjs/website/pull/1129/files#diff-a3f07fb94b91719ead29e589d98c27db44c1f14974480c5960176a2c2c35b6efR56-R66 and the file contents need to be the same, too. I still do not really care how that is achieved, the point is that as long as the bun guide does not lead to the correct setup, then this cannot be merged.

MasedMSD:
I'll get into this page in more detail tomorrow

KnorpelSenf:
Remember to merge `main` in order to fix CI

KnorpelSenf:
Scratch that, it already happened

KnorpelSenf:
I even did it myself \*facepalm\*

KnorpelSenf (commented):
This still does not create the same TS config as `tsc --init` if I understand this correctly

MasedMSD:
Unfortunately we can't fix it. We will have to ask the user to change it manually every time.
But I don't understand the point of this, because everything works with the current tsconfig O.O

KnorpelSenf:
Yeah but I don't want to go through all the code examples and make sure that they compile even with the non-standard TS config that bun introduces. It can be fixed very easily by running `tsc --init` in the getting started guide.

site/docs/guide/getting-started.md
```diff
@@ -150,6 +150,116 @@ in your terminal before you execute `node bot.js`.
...
+```sh
+# Create a new directory and change into it.
+mkdir my-bot
+cd my-bot
+```
+
+Next step you should initialise your project.
+
+```sh
+# Run bun init to scaffold a new project.
```
winstxnhdw, line 170:
```suggestion
# Run bun init to scaffold a new project.
tsc —init
```

winstxnhdw (changes requested):

site/docs/guide/getting-started.md
KnorpelSenf: Nice. Do we also need a `bun add -D typescript` somewhere in there, so that people have `tsc` ready? We might also want to use `bunx`.

MasedMSD: We can't init typescript before `bun init`.
`bun init` creates `package.json` and there could possibly be problems with that.

MasedMSD: We can just use bunx for that btw.
Like

```markdown
# Use bun to init Typescript config
bunx tsc --init
# Run bun init to scaffold a new project.
bun init
```

KnorpelSenf: Right, so adding TypeScript would be done after these two commands? That makes sense

quadratz (changes requested):

site/docs/guide/getting-started.md
quadratz: Wait, `bun init`  already includes a `tsconfig.json`, as explained in https://bun.sh/docs/typescript. I don't think this step is necessary, no?

winstxnhdw: > Wait, `bun init` already includes a `tsconfig.json`, as explained in https://bun.sh/docs/typescript. I don't think this step is necessary, no?

See the earlier comments about keeping tsconfig consistent.

quadratz:
> @KnorpelSenf : Yeah but I don't want to go through all the code examples and make sure that they compile even with the non-standard TS config that bun introduces. It can be fixed very easily by running tsc --init in the getting started guide.
> > @MasedMSD: But I don't understand the point of this, because everything works with the current tsconfig O.O

I am agree with @MasedMSD. I don't see any problem with the [default generated tsconfig.json](https://bun.sh/docs/typescript#suggested-compileroptions). Pretty sure it works well with our examples.

site/docs/guide/introduction.md
```diff
@@ -66,7 +66,7 @@ You will get to know them as you go.
 
 ## Prerequisites to Getting Started
 
-> Skip the rest of this page if you already know how to develop a Deno or a Node.js application, and [get started](./getting-started).
+> Skip the rest of this page if you already know how to develop Deno, Bun or Node.js application, and [get started](./getting-started).
```
quadratz, line 69:
```suggestion
> Skip the rest of this page if you already know how to develop Deno, Bun, or Node.js application, and [get started](./getting-started).
```

site/docs/guide/introduction.md
```diff
@@ -173,6 +173,47 @@ You can stop the bot again with `Ctrl+C`.
 Ready?
 [Get started](./getting-started#getting-started-on-deno)! :robot:
 
+### Prerequisites for Bun
+
+Bun is a new JavaScript runtime, which is another good choice to build bot, like Deno, you are going to write your bot in TypeScript.
```
quadratz, line 178:
```suggestion
Bun is a new JavaScript runtime, which is another good choice to build bot.
Like Deno, you are going to write your bot in TypeScript.
```

quadratz (changes requested):

KnorpelSenf:
> > @KnorpelSenf : Yeah but I don't want to go through all the code examples and make sure that they compile even with the non-standard TS config that bun introduces. It can be fixed very easily by running tsc --init in the getting started guide.
> > > @MasedMSD: But I don't understand the point of this, because everything works with the current tsconfig O.O
> 
> I am agree with @MasedMSD. I don't see any problem with the [default generated tsconfig.json](https://bun.sh/docs/typescript#suggested-compileroptions). Pretty sure it works well with our examples.

Then please go through them and verify this. Also, please test all future changes to the docs for bun compatibility. I don't care what the config file contains, I just won't double the time I spend on code examples just to make a few bun users happy.

MasedMSD:
Fine, so we going to use my [solution](https://github.com/grammyjs/website/pull/1129#discussion_r1829563911)?

KnorpelSenf:
Only as soon as we find somebody who actually puts in the work and makes sure that the code examples work. I don't want to merge docs that don't type-check because then I'm gonna be the one who has to deal with all the complaints.

If somebody is willing to
- make sure that the current docs type-check, and
- make sure that all future docs type-check 

then we can ship docs that use a new TS config setup.

MasedMSD:
@KnorpelSenf It would be useful to create some package for easy unit testing of the bot.
I don't think that's possible given the flood handling by telegram

KnorpelSenf:
Is that in any way related to the current pull request?

MasedMSD:
I mean that we should have some kind of tests to make sure that the bot and the code work. For example, as for the situation now, when we can't be sure that bun will pass all type-checks

winstxnhdw:
> I mean that we should have some kind of tests to make sure that the bot and the code work. For example, as for the situation now, when we can't be sure that bun will pass all type-checks

It’ll be hard to write integration tests for Bun and grammY because the existing test suite is in Deno. We’d have to spend time instrumenting some Docker CI test just specifically for Bun.

KnorpelSenf:
There's no need to actually run any tests, why do you think this? I can't follow

MasedMSD:
@KnorpelSenf still confused. We don't gonna use [this](https://github.com/grammyjs/website/pull/1129#discussion_r1829563911)?

KnorpelSenf:
I don't have strong opinions on what exactly you're gonna do. I just need you to make sure that the code examples are correct. It doesn't matter to me if you achieve this by making people use the setup that is known to work, or if you want to test all present and future code snippets against a new setup.

MasedMSD:
What I mean is that this command generates the typescript config that we use in all the examples. I think we should add Bun, but if there are any problems (which I'm not sure about at all, Bun is quite stable) we should deal with them as they come up.

KnorpelSenf:
So is there still a question? If yes then I don't understand it

MasedMSD:
I mean, this PR has been hanging around for a long time and we need to do something about it. Either put it on hold or get on with it

KnorpelSenf:
I don't care if you use the command or not and I'm not opposed to mentioning bun support

MasedMSD:
> I'm not opposed to mentioning bun support

Great, I'll take care of that PR

KnorpelSenf:
> > I'm not opposed to mentioning bun support
> 
> Great, I'll take care of that PR

Nice

site/docs/guide/files.md
```diff
@@ -181,12 +181,20 @@ new InputFile("/path/to/file");
...
+new InputFile(await Bun.file("/path/to/file").bytes());
+```
+
 :::
 
 #### Uploading Raw Binary Data
 
 You can also send a `Buffer` object, or an iterator that yields `Buffer` objects.
-On Deno, you can send `Blob` objects, too.
+On Deno and Bun you can send `Blob` objects, too.
```
KnorpelSenf, line 197:
```suggestion
On Deno and Bun, you can send `Blob` objects, too.
```

site/docs/guide/files.md
```diff
@@ -181,12 +181,20 @@ new InputFile("/path/to/file");
 new InputFile(await Deno.open("/path/to/file"));
 ```
 
+```ts [Bun]
+// Send a local file.
+new InputFile("/path/to/file");
+
+// Send a `BunFile` instance.
+new InputFile(await Bun.file("/path/to/file").bytes());
```
KnorpelSenf, line 189: Why would
```suggestion
new InputFile(await Bun.file("/path/to/file"));
```
not work?

MasedMSD: Unfortunately no, we will have to fix this at the level of the grammY itself and add support for BunFile.
image

KnorpelSenf: Then their docs are wrong: https://bun.sh/docs/api/file-io#reading-files-bun-file

They say that they conform to `Blob` so that means that `Bun.file("/tmp/file") instanceof Blob` should be `true`.

KnorpelSenf: Their reference is also wrong because their example states that `.bytes()` exists but their reference does not list it. Or is that implicit via `Blob` which they don't seem to inherit? Also, why did they list `.stream()` then?

KnorpelSenf: But well, I guess that you are right that we cannot perform this change

KnorpelSenf:
Closing due to inactivity. I will add bun-specific docs myself for the 2.0 rewrite of the docs.

PR #1131, opened 2024-10-14
Use Deno 2.0 with CI

quadratz:
This should fix the fmt differences between v2 and v1


KnorpelSenf:
Confer #1132

rojvv (approved):
Thank you!

PR #1136, opened 2024-10-22
Update permissions for Deno 2

KnorpelSenf:
grammY now requires `--allow-import` permissions from Deno. This PQ adds that to the docs.

In addition, Deno now runs scripts simply by doing `deno bot.ts` so we migrate to this shorter form, too.

Note for translators: there are only 2 new sentences and 1 edited sentence. They're all in one file. The diff is much larger because I have already updated a number of code snippets for your languages.


rojvv (approved):
Looks good, thank you!

PR #1149, opened 2024-11-12
Fix context code examples

mahovich:
- For `ctx.entities();` there is no key `phone`. The correct key is `phone_number`.
- Added missing `async` to callback function.


KnorpelSenf (approved):
Great catch, thank you so much!

rojvv (approved):

PR #1154, opened 2024-11-17
Translate #1153

KnorpelSenf:
#1153


MasedMSD (approved):

KnorpelSenf:
@rojvv how much time do you currently have to take care of the website? Do you want me to just go ahead for some time and merge things when they are open for days without response?

rojvv:
@KnorpelSenf I have good time for the website, but the only con of it is that it is not consistently consecutive. Feel free to do good things you predict I would be doing.

PR #1157, opened 2024-11-21
Fix Blob instantiations in code examples

MasedMSD:
Changed
```Typescript
const blob = new Blob("ABC", { type: "text/plain" });
```
To
```Typescript
const blob = new Blob(["ABC"], { type: "text/plain" });
```
Due to [blobParts should be iterable](https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob#blobparts)


KnorpelSenf (approved):
TYSM!

rojvv (approved):

PR #1161, opened 2024-11-28
Fix wrong limit for Cloudflare Workers

Chinoman10:
What was stated was the free tier limit, but the limit on the paid plan should be the one stated (_as is the case for the other hosting providers_).


KnorpelSenf (changes requested):
Could you fix the formatting?

KnorpelSenf (approved):
Thank you both!

PR #1164, opened 2024-12-08
Fix typo in adapter of webhookCallback

Hth4nh:
Change "std/http" to "https"

PR #1166, opened 2024-12-11
Use "std/http" adapter for Vercel Edge Functions, instead of "https"

Hth4nh:


KnorpelSenf:
Where do the translations come from? Are you a native speaker in all of these languages?

KnorpelSenf:
Oh. I see. Please revert them. We have a team of translators that will take care of this. It is better to leave translations to them and only submit changes to the English version. 

An exception can be made if a pull request only touches code snippets (such as in #1164) but not here.

Hth4nh:
> Oh. I see. Please revert them. We have a team of translators that will take care of this. It is better to leave translations to them and only submit changes to the English version.
> 
> An exception can be made if a pull request only touches code snippets (such as in #1164) but not here.

Should the outdated line be replaced with an English version so that the translation team can re-translate it later? Or just revert it to its original state?

KnorpelSenf (approved):
Thanks!

I also fixed up a few more things, this is ready for translation now

rojvv (approved):

habemuscode (approved):

habemuscode:
Translations PR: https://github.com/grammyjs/website/pull/1169

This can be merged when reviewed

PR #1168, opened 2024-12-16
Fix code example for similar commands

eikowagenknecht:
The previous example returned both messages when a command was found.


site/docs/plugins/commands.md
```diff
@@ -492,15 +492,15 @@ bot
...
-    // We found a potential match
     if (ctx.commandSuggestion) {
+      // We found a potential match
       await ctx.reply(
         `Hmm... I don't know that command. Did you mean ${ctx.commandSuggestion}?`,
       );
+    } else {
+      // Nothing seems to come close to what the user typed
+      await ctx.reply("Oops... I don't know that command :/");
     }
```
roziscoding, line 503: I'm a strong believer in early return over `else`. Please apply this to the other snippet too :)

```suggestion
    if (ctx.commandSuggestion) {
      // We found a potential match
      return ctx.reply(
        `Hmm... I don't know that command. Did you mean ${ctx.commandSuggestion}?`,
      );
    }
    
    // Nothing seems to come close to what the user typed
    await ctx.reply("Oops... I don't know that command :/");
```

roziscoding (changes requested):

eikowagenknecht:
Changed it to return early in both examples.

roziscoding (approved):

KnorpelSenf (changes requested):
Can you change
```ts
return promise
```
to
```ts
await promise
return
```
in all cases? This implict promise handling is something experienced programmers might do sometimes as a micro optimisation, but it causes a large number of bugs for beginners.

eikowagenknecht:
Sure, I was just going with the suggestion of @roziscoding . I changed to a separate return in the if branch and left out the return at the end of the function now, as it would be meaningless.

KnorpelSenf:
@rojvv I took the liberty to sync everything, we can directly merge into `main` without doing any translations on an intermediate branch

PR #1170, opened 2024-12-19
Add missing word

ikelax:
I think `make` is missing there.


rojvv (approved):
You’re right, thank you.

PR #1172, opened 2024-12-24
Update transformers.md

MobinAskari:
Fix the part where it suggests the wrong way of preventing API calls to Telegram

### DIFF
--- a/site/docs/advanced/transformers.md
+++ b/site/docs/advanced/transformers.md
@@ -35,8 +35,8 @@ bot.use((ctx, next) => next());
 Here is an example of a transformer function that prevents all API calls from happening:
 
 ```ts
-// Incorrectly return undefined instead of the respective object types.
-bot.api.config.use((prev, method, payload) => undefined as any);
+// Incorrectly return `{ ok: true } as any` instead of the respective object types.
+return { ok: true } as any
 ```
 
 You can also install transformer functions on the context object's API object.


site/docs/advanced/transformers.md
```diff
@@ -35,8 +35,8 @@ bot.use((ctx, next) => next());
 Here is an example of a transformer function that prevents all API calls from happening:
 
 ```ts
-// Incorrectly return undefined instead of the respective object types.
-bot.api.config.use((prev, method, payload) => undefined as any);
+// Incorrectly return `{ ok: true } as any` instead of the respective object types.
+return { ok: true } as any
```
KnorpelSenf, line 39:
```suggestion
bot.api.config.use((prev, method, payload) => { ok: true } as any);
```

site/docs/advanced/transformers.md
```diff
@@ -35,8 +35,8 @@ bot.use((ctx, next) => next());
 Here is an example of a transformer function that prevents all API calls from happening:
 
 ```ts
-// Incorrectly return undefined instead of the respective object types.
-bot.api.config.use((prev, method, payload) => undefined as any);
+// Incorrectly return `{ ok: true } as any` instead of the respective object types.
+bot.api.config.use((prev, method, payload, signal) => { ok: true } as any);
```
KnorpelSenf, line 39: I'm so dumb
```suggestion
bot.api.config.use((prev, method, payload, signal) => ({ ok: true }) as any);
```

MobinAskari:
I've got tunnel vision man...
I literally don't know why I only wrote the return statement but let's not talk about it, that was THE embarrassing part lol.

PR #1173, opened 2024-12-24
Fix return type (#1172)

KnorpelSenf:


rojvv (approved):

LWJerri (approved):

site/docs/advanced/transformers.md
```diff
@@ -35,8 +35,8 @@ bot.use((ctx, next) => next());
 Here is an example of a transformer function that prevents all API calls from happening:
 
 ```ts
-// Incorrectly return undefined instead of the respective object types.
-bot.api.config.use((prev, method, payload) => undefined as any);
+// Incorrectly return `{ ok: true } as any` instead of the respective object types.
+bot.api.config.use((prev, method, payload, signal) => ({ ok: true }) as any);
```
KnorpelSenf, line 39: Bracket placement is off. Are we sure that this code snippet parses correctly and is touched by `deno fmt`?

PR #1183, opened 2025-01-24
Merge fmt and lint CI jobs

KnorpelSenf:
Both jobs take a few seconds to set up and a few milliseconds to complete. The setup of these jobs is identical. It makes more sense to merge them.


.github/workflows/check.yml
```diff
@@ -7,7 +7,7 @@ on:
     types: ["opened", "edited", "synchronize"]
 
 jobs:
-  fmt:
+  format-and-lint:
```
rojvv, line 10:
```suggestion
  fmt-lint:
```

rojvv (approved):

PR #1188, opened 2025-02-09
Add warning to conversations migration instructions

roziscoding:


site/docs/plugins/conversations.md
```diff
@@ -1497,6 +1498,12 @@ await conversation.external((ctx) => { // [!code ++]
 > Accessing `ctx.session` was possible with 1.x, but it was always incorrect.
 > `ctx.session` is no longer available with 2.x.
 
+::: warning Sending data back to the outside context
```
KnorpelSenf, line 1501:
```suggestion
::: warning Sending Data Back to the Outside Context
```

KnorpelSenf (changes requested):
I had to read this thrice before I even understood what this means. Can you add a code example illustrating good and bad usage (similar to the existing code examples) in order to make this a little bit more clear?

rojvv (approved):

KnorpelSenf (changes requested):
I'd still like to see a code example

roziscoding:
> I'd still like to see a code example

I'll work on it soon™️

roziscoding:
I don't think I'll pick this back up. If anyone else wants to do so, feel free. If not, we can close this.

PR #1190, opened 2025-02-11
chore: Explain how to use `Promise.race` in conversations

LWJerri:
It all started from https://t.me/grammyjs_ru/80756

### DIFF
--- a/site/docs/plugins/conversations.md
+++ b/site/docs/plugins/conversations.md
@@ -1256,6 +1256,44 @@ Internally, when several wait calls are reached at the same time, the conversati
 As soon as the next update arrives, it will then replay the conversation builder function once for each encountered wait call until one of them accepts the update.
 Only if none of the pending wait calls accepts the update, the update will be dropped.
 
+## Waiting for One of Many Wait Calls
+
+You can wait for multiple of wait calls simultaneously using the [Promise.race](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) method.
+This allows you to handle different responses in a single conversation flow.
+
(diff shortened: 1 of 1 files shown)


site/docs/plugins/conversations.md
```diff
@@ -1256,6 +1256,44 @@ Internally, when several wait calls are reached at the same time, the conversati
 As soon as the next update arrives, it will then replay the conversation builder function once for each encountered wait call until one of them accepts the update.
 Only if none of the pending wait calls accepts the update, the update will be dropped.
 
+## Waiting for One of Many Wait Calls
+
+You can wait for multiple of wait calls simultaneously using the [Promise.race](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) method.
```
KnorpelSenf, line 1261: This is incorrect. Using `Promise.race` is a bad idea and will behave unexpectedly. It will resolve with the first update, but then it will swallow all other updates until all wait calls have resolved. These updates are lost forever (simply due to how `Promise.race` works) so it is never useful.

KnorpelSenf (changes requested):
There currently is no way to disable wait calls after they have been reached.

You can technically work around the issue by doing something like

```ts
let resolved = false
const first = conversation.waitUntil(cond0).and(() => !resolved).finally(() => resolved = true)
const second = conversation.waitUntil(cond1).and(() => !resolved).finally(() => resolved = true)
const ctx = await Promise.race([first, second])
```
which would effectively disable one of the wait calls as soon as the other has resolved. However, we decided against adding this as the default behaviour to the plugin because it can seem quite magical.

PR #1196, opened 2025-02-14
update thank you component to use UTC date

roziscoding:
This makes it so the thanked person is synced across the globe.
I'ts true that it might change during the day for people who are not on UTC timezones, but everyone will still be featured for 24 hours.


KnorpelSenf (approved):
Great fix

PR #1201, opened 2025-03-16
Added a note that GrammY does not implement editing of Telegram Busin…

rtyt3000:
…ess messages 
This pull request includes updates to the documentation regarding the support for business messages in grammY. The changes clarify the current limitations of grammY concerning message editing and deletion in business chats.


KnorpelSenf:
You may have misunderstood me, this is implemented in grammY. There is no such limitation. Look [here](https://github.com/grammyjs/grammY/blob/d9064bca94df37b5264ef5f1bcb0b9e02f662786/src/context.ts#L2610).

The docs say at https://grammy.dev/advanced/business#editing-or-deleting-business-messages that business messages cannot be edited by bots. This is the part that is incorrect. The docs need to be updated.

rtyt3000:
The fact of the matter is that if implemented, it is very strange, at least conventional methods will write simply Message not found

KnorpelSenf:
Hmm perhaps there's a bug then, let me take a look at this later today

KnorpelSenf:
@rtyt3000 can you try out the changes in https://github.com/grammyjs/grammY/pull/775 and confirm that they work? You can install them via `npm i github:grammyjs/grammY#fix-biz-id`. If that fixes it, please leave an approval on https://github.com/grammyjs/grammY/pull/775.

PR #1205, opened 2025-03-30
Fix parameters passed to setMessageReaction

dreminjs:


KnorpelSenf:
@dreminjs what makes you think that this method is deprecated?

dreminjs:
Because it doesnt work :) > @dreminjs what makes you think that this method is deprecated?

dreminjs:
Because it doesnt work

KnorpelSenf:
@dreminjs nice catch. It's actually not deprecated, it just never existed. This was a typo in the docs.

Can you sync your changes to the other languages, as well? :)

KnorpelSenf (approved):
LGTM, thank you so much!

rojvv (approved):

PR #1214, opened 2025-05-14
Update Cloudflare hosting guides to use global env

### LINKED ISSUE #1213: Update the Cloudflare-related hosting guides
Since March 17 2025, workers can access environment variables and secrets from the global scope.
https://developers.cloudflare.com/changelog/2025-03-17-importable-env

Now, we can set up bots outside the `fetch` method, which is a much better approach than before.

niusia-ua:
Closes #1213.

I have updated all Cloudflare hosting guides for Node.js.
I haven't touched Denoflare guides because I don't know how it affects them.


KnorpelSenf (approved):
Great stuff

PR #1218, opened 2025-06-16
Replace npx with pnpm exec

icy-comet:


rojvv (commented):
Can you update it in other places as well?

PR #1221, opened 2025-07-18
Improve API flavors example

evermake:


evermake:
> So the `SomeContextFlavor` in this case could be something like this
> 
> ```ts
> export type SomeContextFlavor = {
>   someMethod: (...) => Promise<...>
> }
> ```
> 
> right?

@KnorpelSenf, could you answer this, please?

KnorpelSenf:
@evermake that would be the API flavour, but yes, that structure is correct. For the context flavour, it needs to adjust `ctx.api` etc.

evermake:
@KnorpelSenf, should I update labels somehow to make CI pass?

KnorpelSenf:
Nope, we need to wait for translations. Once they're done (all language labels are added) the job will pass.

PR #1240, opened 2025-09-30
Resolve Hostinger redirections

KnorpelSenf:
It redirects inconsistently. Perhaps this root URL is never redirected?


KnorpelSenf:
That's why only those links are updated here. 

The root URL doesn't always redirect me for some reason. In some cases, it just accepts a plain `/` path. I have no idea why.

KnorpelSenf:
Oh and in some cases, it redirects the other way. Look at the current CI run. How to proceed?

PR #1284, opened 2026-03-05
fix(by copilot): stop escaping HTML in JSDoc strings

### LINKED ISSUE #1027: Broken link in core API reference
https://grammy.dev/ref/core/Context#getfile

KnightNiwrem:
LLM disclosure
Harness: Github Copilot CLI
Model: Claude Opus 4.6 (reasoning: high)

Remove the while loop in P.tsx that escaped <...> patterns to &lt;...&gt; in JSDoc text. This was breaking intentional HTML tags (like links) in JSDoc strings, causing broken links in the API reference.

Fixes #1027

### DIFF
--- a/site/api/components/P.tsx
+++ b/site/api/components/P.tsx
@@ -11,26 +11,7 @@ export function P(
   },
 ) {
   if (props.doc && props.children) {
-    const parts = props.children.split(/(```|```ts)/);
-    const newParts = new Array<string>();
-    let inCodeBlock = false;
-    for (let part of parts) {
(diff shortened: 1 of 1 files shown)


KnorpelSenf (changes requested):
This breaks the build for obvious reasons

KnorpelSenf:
Does this actually still happen? A lot more parsers worked like ours, so I fixed most issues in the TSDoc comments directly. At this point, I am no longer sure if our parser is actually as broken as I thought.

I think we'll abandon this. Breaks the build anyway.

PR #1292, opened 2026-04-17
docs: note that lazy sessions and multi sessions are mutually exclusive

### LINKED ISSUE #936: Missing documentation about multi lazy sessions
**ORIGINAL COMMENT**

              Hang on, is this only breaking when you're using lazy multi sessions? If so, then the problem is that there are no lazy multi sessions. This isn't supported. We implemented it initially, but combining lazy sessions with multi sessions leads to numerous strange edge cases that are hard to understand and by design impossible to mitigate, so we dropped this again. If you were using TypeScript, your editor could've told you this. Since you don't seem to be using it, here is a link to the API reference of grammY which tells you this: https://deno.land/x/grammy@v1.12.4/mod.ts?s=lazySession

I'll add another example for this to the docs in the coming days. There is no bug here.

I will close this issue. Feel free to comment or reopen if you have further questions :)

_Originally posted by @KnorpelSenf in https://github.com/grammyjs/grammY/issues/345#issuecomment-1366223793_
            

**PROBLEM**

About the cited comment, it would be nice to have it written in the documentation [here](https://grammy.dev/plugins/session#multi-sessions) and maybe also a possible workaround example

Dramex:
Closes #936.

The Lazy Sessions section reads as if the middleware has feature parity with strict sessions, but in fact \`lazySession\` does not accept a \`type: \"multi\"\` configuration — on the thread the maintainer confirmed this and that there is no workaround. Until this is documented, users assume they can combine the two and are surprised when it fails.

Adds a short warning callout at the top of the Lazy Sessions section stating the constraint and pointing readers who need multiple session namespaces back to the default strict \`session\` middleware with \`type: \"multi\"\`.

No code changes — docs only, English-language only for now (translations can follow per the usual process).


site/docs/plugins/session.md
```diff
@@ -641,6 +641,12 @@ You can then keep using `SessionFlavor<SessionData>` for your context object.
 
 Lazy sessions is an alternative implementation of sessions that can significantly reduce the database traffic of your bot by skipping superfluous read and write operations.
 
+::: warning Lazy sessions do not support multi sessions
+Lazy sessions and [multi sessions](#multi-sessions) are mutually exclusive.
+`lazySession` is a separate implementation of the session middleware, and it does not accept the `type: "multi"` configuration that the strict `session` middleware does.
+If you need several independent session namespaces (for example, one per chat and one per user), use the default strict `session` middleware with `type: "multi"`.
+:::
```
KnorpelSenf, line 638: This takes up a bit too much space IMO. What do you think of this?
```suggestion
> Note that the lazy session implementation has no concept of multi sessions.
> Stick with the regular session implementation if you want to use its [multi sessions](#multi-sessions).
```

KnorpelSenf:
@Dramex are you still interested in getting this merged?

PR #1293, opened 2026-04-17
fix: escape each angle-bracket placeholder in ref JSDoc independently

### LINKED ISSUE #1027: Broken link in core API reference

Dramex:
Closes #1027.

The reporter's screenshot shows a broken render on \`/ref/core/Context#getfile\`. The \`getFile\` JSDoc says:

> ...the link \`https://api.telegram.org/file/bot<token>/<file_path>\`, where \`<file_path>\` is taken from the response...

The ref generator tries to escape those angle brackets in \`site/api/components/P.tsx\`:

\`\`\`ts
while (/<.+>/.test(part)) {
    part = part.replace(/<(.+)>/, \"&lt;\$1&gt;\");
}
\`\`\`

The regex is greedy, so on that line the first iteration matches from the **first** \`<\` to the **last** \`>\` — everything in between becomes the capture, including the \`>/<\` that separates the two placeholders. After one iteration you get \`bot&lt;token>/<file_path&gt;\`, the \`.test\` no longer matches (there is no trailing \`>\` after the remaining \`<\`), and the loop exits with broken HTML that the browser then parses as a partial tag.

Fix by replacing the loop with a single \`replace\` using \`[^<>]+\` inside the capture group and the \`g\` flag. Each placeholder is escaped on its own, there is no backtracking across other placeholders, and the loop is no longer needed:

\`\`\`ts
part = part.replace(/<([^<>]+)>/g, \"&lt;\$1&gt;\");
\`\`\`

Added a code comment linking back to #1027 so future readers know why \`[^<>]+\` was chosen.

## Test plan

Quick verification in a REPL:

\`\`\`js
\"bot<token>/<file_path>\".replace(/<([^<>]+)>/g, \"&lt;\$1&gt;\")
// \"bot&lt;token&gt;/&lt;file_path&gt;\"  ✓
\`\`\`

Also tested a single-placeholder line (\`foo<bar>\` → \`foo&lt;bar&gt;\`) and a line with no placeholders (unchanged), both behave the same as before. Build would need to run on Netlify to regenerate ref pages — I did not deploy locally.


KnorpelSenf:
Does this actually still happen? A lot more parsers worked like ours, so I fixed most issues in the TSDoc comments directly. At this point, I am no longer sure if our parser is actually as broken as I thought.

KnorpelSenf:
Cannot reproduce. Closing. Please re-open if this is still needed.

PR #1301, opened 2026-06-18
fix: use correct apostrophe

LWJerri:


KnorpelSenf (commented):
Isn't this done automatically during the build step because we have enabled smart quotes?

If no, would it be better to fix vitepress and add support there? If yes, we can merge this in the meantime and then revert later on.

LWJerri:
> Isn't this done automatically during the build step because we have enabled smart quotes?
> 
> If no, would it be better to fix vitepress and add support there? If yes, we can merge this in the meantime and then revert later on.

Tbh, I don't know what "smart quotes" are, but it seems like this feature isn't enabled, and I'm not sure if it can even be enabled, since we haven't received any notifications over all this time indicating that we have incorrect quotation marks in some places.

KnorpelSenf:
I was referring to the `typographer` config option which is set here: https://github.com/grammyjs/website/blob/68a3f373238f09ad4a3c3ad86fb31fccdbe89c75/site/docs/.vitepress/config.ts#L41

Look for smartquotes here: https://markdown-it.github.io/markdown-it/?hl=de-DE

KnorpelSenf (approved):
Understood

PR #1305, opened 2026-07-15
docs: add Editing Messages

gameroman:


site/docs/guide/basics.md
```diff
@@ -131,3 +131,17 @@ bot.command("start", async (ctx) => {
...
+
+You can edit a message sent by your bot like this:
+
+```ts
+bot.command("think", async (ctx) => {
+  const msg = await ctx.reply("Thinking...");
+
+  const result = await getSomeResult();
+
+  await ctx.api.editMessageText(ctx.chatId, msg.message_id, `The result is ${result}.`);
```
KnorpelSenf: I guess this should just be: 
```suggestion
  await ctx.editMessageText(`The result is ${result}.`);
```
Will also appear in auto-complete if you just type `ctx.edit|`. Still worth adding this?

gameroman: `ctx.editMessageText` will not work since it will try to edit the user's message, not the bot's message

https://t.me/grammyjs/366175

KnorpelSenf: Oh right, I somehow thought of callback queries not commands

gameroman:
```suggestion
  const newText = await getSomeNewText();

  await ctx.api.editMessageText(ctx.chatId, msg.message_id, newText);
```

KnorpelSenf (changes requested):
@gameroman I think we should add a link to the hydrate plugin here. It prevents the exact confusion that got me. Can you add it? LGTM otherwise, thanks for adding this

gameroman:
> hydrate plugin

This one https://grammy.dev/plugins/hydrate.html ?

KnorpelSenf:
Yep! No need for a source example here, just a hint that working with message objects (like editing them) is simplified by the plugin.

site/docs/guide/basics.md
```diff
@@ -131,3 +131,17 @@ bot.command("start", async (ctx) => {
...
+
+```ts
+bot.command("think", async (ctx) => {
+  const msg = await ctx.reply("Thinking...");
+
+  const newText = await getSomeNewText();
+
+  await ctx.api.editMessageText(ctx.chatId, msg.message_id, newText);
+});
+```
```
gameroman, line 147: Something like this?

````suggestion
```

You might also want to consider the [hydration plugin](https://grammy.dev/plugins/hydrate) that make this simpler.
````

Comment (KnorpelSenf) [reply site/docs/guide/basics.md:147]: Yes:
````suggestion
```

> Consider using the [hydrate plugin](../plugins/hydrate) which simplifies message editing.
````

KnorpelSenf (approved):
TYSM!

KnorpelSenf:
The changes will be translated on #1308. We are going to publish them as soon as that's done. @gameroman there's nothing else you need to do here, just letting you know :)

